PR #1, opened 2020-06-11
AD authentication changes

ajelenak:


src/rest_vol.c
```diff
@@ -687,6 +727,7 @@ H5_rest_set_connection_information(void)
          * URL, username and password key-value pairs.
          */
         while (fgets(file_line, sizeof(file_line), config_file) != NULL) {
+            if ((file_line[0] == '#') || !strlen(file_line)) continue;
```
jhendersonHDF, line 730: This line should probably be moved below the variable declarations and then have have the variables get initialized after this line. Otherwise, compiling in ISO C mode will complain about variable declarations after a statement.

src/rest_vol.c
```diff
@@ -732,6 +796,350 @@ H5_rest_set_connection_information(void)
...
+            FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTGET, FAIL, "reading access token config file - unable to retrieve location of home directory");
+
+        if (NULL == (home_dir = getenv("HOMEPATH")))
+            FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTGET, FAIL, "reading access token config file - unable to retrieve location of home directory");
+
+        token_cfg_file_pathname_len = strlen(home_drive) + strlen(home_dir) + strlen(token_cfg_file_name) + 3;
+        if (NULL == (token_cfg_file_pathname = RV_malloc(token_cfg_file_pathname_len)))
+            FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTALLOC, FAIL, "unable to allocate space for access token config file pathname");
+
+        if (snprintf(token_cfg_file_pathname, token_cfg_file_pathname_len, "%s\%s\%s", home_drive, home_dir, token_cfg_file_name) < 0)
```
jhendersonHDF, line 867: Looks like I left a missing extra '\' character to escape the '\' in the output string. I know this same code exists in the .hscfg file parsing logic, so I wonder how that works correctly for Chevron on Windows..

src/rest_vol.c
```diff
@@ -732,6 +796,350 @@ H5_rest_set_connection_information(void)
...
+            CURL_PERFORM(curl, H5E_VOL, H5E_CANTGET, FAIL);
+
+            yajl_tree_free(parse_tree);
+            parse_tree = NULL;
+
+#ifdef RV_CONNECTOR_DEBUG
+            printf("-> Authentication successfully completed.\n");
+            printf("-> Authentication server response:\n-> %s\n", response_buffer.buffer);
+#endif
+
```
jhendersonHDF, line 1077: It seems like everything below here is probably applicable to the unattended authentication case as well isn't it? Should it be moved outside of this block?

src/rest_vol.c
ajelenak: Why not move them all the way "up" at the beginning of the `else` block? I wanted to avoid unnecessary parsing of lines with no interesting content.

jhendersonHDF: Sure, I understand that it's to avoid parsing any lines that are just comments. I'm just saying it should go after key and val are declared or C89 mode for C compilers will complain. Basically, just change this to:

const char *key;
const char *val;

if ((file_line[0] == '#') || !strlen(file_line)) continue;

key = strtok(file_line, " =\n");
val = strtok(NULL, " =\n");

That will satisfy ISO C compiling mode and also avoid tokenizing the line until after it's been checked for being a comment line. The only gotcha may be that some compilers might complain about not initializing a const variable and assigning to it later. You may need to remove the const in this case if so.

ajelenak: Done.

src/rest_vol.c
ajelenak: Fixed in both places.

src/rest_vol.c
ajelenak: Moved the first line out so it applies to both cases.

src/rest_vol.c
jhendersonHDF: Great, thanks!

src/rest_vol.c
jhendersonHDF: Ah, maybe I'm wrong, but I meant that it seems like all the code for checking the validity of the access token and also retrieving the refresh token would also apply to the unattended case, but currently it looks like this only applies for the attended case's code. I'm guessing you'll probably have to move some stuff around when doing the token refreshing work anyway though, so this will work for now.

PR #3, opened 2020-08-04
Custom user agent

ajelenak:
Since this VOL is a web client I thought it should have a custom user agent to claim its rightful place in the World Wide Web: Below is the current format:
```
User-Agent: libhdf5/1.12.0 (x86_64-apple-darwin17.7.0) VOL-REST/1.0.0 libcurl/7.71.1 libyajl/2.1.0
```
I'm not sure the VOL identifier is the best possible yet, and whether to keep libyajl, but it's a start.

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -342,6 +342,8 @@ static herr_t
 H5_rest_init(hid_t vipl_id)
 {
     herr_t ret_value = SUCCEED;
+    curl_version_info_data *curl_ver = NULL;
+    char user_agent[128];
 
     /* Check if already initialized */
     if (H5_rest_initialized_g)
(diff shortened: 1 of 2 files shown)


jreadey:
I guess I should add a user agent for h5pyd as well.

How are/should we be managing the REST_VOL versions?  I assume 1.0.0 is what we had before the 1.12 updates and AD support.

Do we expect further changes in master to support older hdf5 lib versions?  Dealing with different versions in different branches sounds problematic.

jhendersonHDF:
I think it's definitely useful to have a user agent for the VOL. I don't really have any input on whether we should keep libyajl, other than saying that it's not really related to any web API format since it's just a JSON library.

> How are/should we be managing the REST_VOL versions? I assume 1.0.0 is what we had before the 1.12 updates and AD support.
> 
> Do we expect further changes in master to support older hdf5 lib versions? Dealing with different versions in different branches sounds problematic.

I guess we haven't really had an official 1.0.0 release yet (nothing's been tagged as such), so that number is essentially meaningless until that point. HDF5 is still changing as well due to issues in the VOL layer, so there isn't really a good concrete release to base the versioning off of yet either. It would be convenient if the REST VOL versions could nicely track alongside HDF5 releases, but this often ends up being problematic.

For the REST VOL, I suppose the major version should change whenever a breaking API change is introduced/we move to a new major HDF5 release, the minor version should change whenever we move to a new minor HDF5 release, and then the release version should change as larger updates are made to the connector. We could also just reserve the major/minor/release versions solely for the connector and then add the required HDF5 library version as part of the release tag name, e.g. VOL-REST-1.0.0-hdf5-1-12

ajelenak:
I think that any VOL plugin should have a separate versioning from the HDF5 library.

Another issue: The name of this VOL is generic. It would be better to call it "HDF5 REST" because that's the specific RESTful API it supports. I hope there will be more VOL plugins for other REST APIs.

And thinking of the user agent today, I think a more succinct content is better. Something like:
```
User-Agent: libhdf5/1.12.0 (x86_64-apple-darwin17.7.0) HDF5_REST_VOL/1.0.0
```

ajelenak:
This is the final version for now:
```
User-Agent: libhdf5/1.12.0 (x86_64-apple-darwin17.7.0; HDF5 REST VOL v1.0.0)
```

PR #4, opened 2020-11-02
added option to link against static thread safe hdf5 library

jpswinski:
Adds option to link against thread-safe build of hdf5 library.  The option is specified by 's' to the build_vol_cmake.sh script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -398,6 +398,15 @@ if (HDF5_VOL_REST_ENABLE_COVERAGE)
 endif ()
 
 #-----------------------------------------------------------------------------
+# Option to use code coverage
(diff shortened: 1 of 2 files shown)


jhendersonHDF (approved):
Looks good to me. Is there a reason that HDF5 would have to be a static library in this case? I only ask because in order to use HDF5's dynamic loading feature for the connector, one would need to build with a shared library version of HDF5. I don't know if that's a major concern for you or not though.

jpswinski:
@jhendersonHDF the shared library may work without needing the above changes, it may just pull in the pthread library automatically; I'd have to test it out.  We were using the static version to simplify some of our install scripts downstream so I didn't try the shared library version.

PR #10, opened 2023-02-28
Update REST VOL to be compatible with HDF 1.14

mattjala:
Updated callback signatures
Removed references to deprecated macros
Assigned the REST VOL the official identifier 520
Updated `H5VL_class_t` struct to `HDF5_VOL_REST_VERSION` 3

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -33,11 +33,10 @@
  * its responses into
  */
 #define CURL_RESPONSE_BUFFER_DEFAULT_SIZE             1024
-
 /*
  * The VOL connector identification number.
  */
-hid_t H5_rest_id_g = H5I_INVALID_HID;
+hid_t H5_rest_id_g = HDF5_VOL_REST_CLS_VAL;
 
 static hbool_t H5_rest_initialized_g = FALSE;
 
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -125,7 +125,7 @@ static herr_t H5_rest_authenticate_with_AD(H5_rest_ad_info_t *ad_info);
 
 /* Introspection callbacks */
 static herr_t H5_rest_get_conn_cls(void *obj, H5VL_get_conn_lvl_t lvl, const struct H5VL_class_t **conn_cls);
-static herr_t H5_rest_opt_query(void *obj, H5VL_subclass_t cls, int opt_type, hbool_t *supported);
+static herr_t H5_rest_opt_query(void *obj, H5VL_subclass_t cls, int opt_type, uint64_t *flags);
 
 /* cURL function callbacks */
 static size_t H5_rest_curl_read_data_callback(char *buffer, size_t size, size_t nmemb, void *inptr);
@@ -1502,15 +1502,15 @@ done:
  *---------------------------------------------------------------------------
  */
 static herr_t
-H5_rest_opt_query(void *obj, H5VL_subclass_t cls, int opt_type, hbool_t *supported)
+H5_rest_opt_query(void *obj, H5VL_subclass_t cls, int opt_type, uint64_t *flags)
 {
     herr_t ret_value = SUCCEED;
 
-    if (!supported)
-        FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "\"supported\" is NULL");
+    if (!flags)
+        FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "\"flags\" is NULL");
 
     /* Advertise no current support for any optional operations that are queried. */
-    *supported = FALSE;
+    *flags = FALSE;
 
 done:
     return ret_value;
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -2418,7 +2418,7 @@ RV_convert_dataspace_shape_to_JSON(hid_t space_id, char **shape_body, char **max
                     CHECKED_REALLOC(shape_out_string, shape_out_string_curr_len, (size_t) buf_ptrdiff + MAX_NUM_LENGTH + 1,
                                     shape_out_string_curr_pos, H5E_DATASPACE, FAIL);
 
-                    if ((bytes_printed = sprintf(shape_out_string_curr_pos, "%s%llu", i > 0 ? "," : "", dims[i])) < 0)
+                    if ((bytes_printed = sprintf(shape_out_string_curr_pos, "%s%llu", i > 0 ? "," : "", (long long unsigned int) dims[i])) < 0)
                         FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_SYSERRSTR, FAIL, "sprintf error");
                     shape_out_string_curr_pos += bytes_printed;
                 } /* end if */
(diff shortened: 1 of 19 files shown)


src/rest_vol_file.c
```diff
@@ -539,14 +536,18 @@ RV_file_specific(void *obj, H5VL_file_specific_t specific_type, hid_t dxpl_id,
...
             FUNC_GOTO_ERROR(H5E_FILE, H5E_UNSUPPORTED, FAIL, "H5Fmount is unsupported");
             break;
+        */
 
         /* H5Funmount */
+        /*
         case H5VL_FILE_UNMOUNT:
             FUNC_GOTO_ERROR(H5E_FILE, H5E_UNSUPPORTED, FAIL, "H5Funmount is unsupported");
             break;
+        */
```
jhendersonHDF, line 550: We should go ahead and remove these blocks for H5Fmount and H5Funmount since they're handled elsewhere now and also not used very much, if at all.

src/rest_vol_link.h
```diff
@@ -18,19 +18,18 @@ extern "C" {
 #endif
 
 /* REST VOL Link callbacks */
-herr_t RV_link_create(H5VL_link_create_type_t create_type, void *obj,
-    const H5VL_loc_params_t *loc_params, hid_t lcpl_id, hid_t lapl_id,
-    hid_t dxpl_id, void **req, va_list arguments);
+herr_t RV_link_create(H5VL_link_create_args_t *args, void *obj, const H5VL_loc_params_t *loc_params,
+                  const H5VL_class_t *cls, hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req);
```
jhendersonHDF, line 22: I believe this prototype is mismatched. Shouldn't have an `H5VL_class_t` parameter.

src/rest_vol_attr.c
```diff
@@ -114,7 +114,7 @@ RV_attr_create(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_
 
     /* If this is a call to H5Acreate_by_name, locate the real parent object */
     if (H5VL_OBJECT_BY_NAME == loc_params->type) {
-        htri_t search_ret;
+        hbool_t search_ret;
```
jhendersonHDF, line 117: Since RV_find_object_by_path doesn't appear to have changed, all of these places where `htri_t search_ret` was changed to `hbool_t search_ret` should just remain as `htri_t search_ret` to match the prototype for RV_find_object_by_path. That, or RV_find_object_by_path should be changed to return an herr_t so we can detect failure, and then accept an `hbool_t` parameter through which it returns the existence status of the object being looked for. In the "link exists" and "attribute exists" callbacks, those arguments being passed in DO need to change from `htri_t` to `hbool_t`, but the `search_ret` variables in those callbacks still need to be `htri_t`. Eventually htri_t will probably go away, but for now

src/rest_vol_link.c
```diff
@@ -73,8 +73,11 @@ const char *link_collection_keys2[] = { "collection", (const char *) 0 };
  *              July, 2017
  */
 herr_t
-RV_link_create(H5VL_link_create_type_t create_type, void *obj, const H5VL_loc_params_t *loc_params,
-    hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req, va_list arguments)
+RV_link_create(H5VL_link_create_args_t *args, void *obj, const H5VL_loc_params_t *loc_params,
+                  const H5VL_class_t *cls, hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req)
+
+/* (H5VL_link_create_t create_type, void *obj, const H5VL_loc_params_t *loc_params,
+    hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req, va_list arguments) */
```
jhendersonHDF, line 80: There's a few of these commented out old prototypes that we should just remove now.

src/rest_vol_dataset.c
```diff
@@ -389,15 +389,15 @@ RV_dataset_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name
...
+                   hid_t file_space_id[], hid_t dxpl_id, void *buf[], void **req)
 {
     H5S_sel_type  sel_type = H5S_SEL_ALL;
-    RV_object_t  *dataset = (RV_object_t *) obj;
+    RV_object_t  *dataset = (RV_object_t *) dset[0];
     H5T_class_t   dtype_class;
     hssize_t      mem_select_npoints, file_select_npoints;
     hbool_t       is_transfer_binary = FALSE;
-    htri_t        is_variable_str;
+    hbool_t        is_variable_str;
```
jhendersonHDF, line 400: Same here for the `htri_t` -> `hbool_t` issue. Apologies if I wasn't very clear on that; `htri_t` is still used in different places in the library, we just changed the two arguments for the "link exists" and "attribute exists" callbacks from `htri_t` -> `hbool_t`.

src/rest_vol.c
```diff
@@ -293,25 +294,25 @@ H5rest_init(void)
     if (H5open() < 0)
         FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTINIT, FAIL, "HDF5 failed to initialize");
 
-    if (H5_rest_id_g >= 0 && (idType = H5Iget_type(H5_rest_id_g)) < 0)
```
jhendersonHDF, line 296: This line should probably stay here. The VOL connector could possibly be initialized by env. variable rather than H5rest_init, so this line attempts to check if the connector is already registered before doing it again. We could keep a second copy inside the `H5I_VOL != idType` block if we want to be extra sure, but this first instance of it is useful.

src/rest_vol.c
```diff
@@ -2414,7 +2418,7 @@ RV_convert_dataspace_shape_to_JSON(hid_t space_id, char **shape_body, char **max
                     CHECKED_REALLOC(shape_out_string, shape_out_string_curr_len, (size_t) buf_ptrdiff + MAX_NUM_LENGTH + 1,
                                     shape_out_string_curr_pos, H5E_DATASPACE, FAIL);
 
-                    if ((bytes_printed = sprintf(shape_out_string_curr_pos, "%s%llu", i > 0 ? "," : "", dims[i])) < 0)
+                    if ((bytes_printed = sprintf(shape_out_string_curr_pos, "%s%llu", i > 0 ? "," : "", (long long unsigned int) dims[i])) < 0)
```
jhendersonHDF, line 2421: Rather than converting these to `long long`, we actually have a few printf macro for `hsize_t` values in H5public.h such as `PRIuHSIZE`. So the format here could be `"%s" PRIuHSIZE`.

src/rest_vol_link.c
```diff
@@ -127,11 +127,11 @@ RV_link_create(H5VL_link_create_type_t create_type, void *obj, const H5VL_loc_pa
     if (!(new_link_loc_obj->domain->u.file.intent & H5F_ACC_RDWR))
         FUNC_GOTO_ERROR(H5E_FILE, H5E_BADVALUE, FAIL, "no write intent on file");
 
-    switch (create_type) {
+    switch (args->op_type) {
         /* H5Lcreate_hard */
         case H5VL_LINK_CREATE_HARD:
         {
-            htri_t  search_ret;
+            hbool_t  search_ret;
```
jhendersonHDF, line 134: Six remaining `search_ret` reversions for this file

src/rest_vol_dataset.c
```diff
@@ -288,7 +288,7 @@ RV_dataset_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name
     RV_object_t *parent = (RV_object_t *) obj;
     RV_object_t *dataset = NULL;
     H5I_type_t   obj_type = H5I_UNINIT;
-    htri_t       search_ret;
+    hbool_t       search_ret;
```
jhendersonHDF, line 291: Four remaining `hbool_t` -> `htri_t` reversions for this file, including `search_ret` and `is_variable_str`

jhendersonHDF:
Just a couple minor cleanup things left, but everything is looking good in this PR

jhendersonHDF (approved):

PR #12, opened 2023-03-13
Add autotools build option to connect to socket

mattjala:
Also added a workflow to run the REST VOL tests using the socket on push, best-guess capability flags for the connector, and replaced `sprintf` -> `snprintf`.

Currently the socket option is hardcoded to use `/tmp/hs/sn_1.sock`

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -395,13 +395,14 @@ H5_rest_init(hid_t vipl_id)
     if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_USERAGENT, user_agent))
         FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "error while setting CURL option (CURLOPT_USERAGENT)");
 
+    const char *URL = getenv("HSDS_ENDPOINT");
+    
+    if (!strncmp(URL, UNIX_SOCKET_PREFIX, strlen(UNIX_SOCKET_PREFIX))) {
+        char* socket_path = "/tmp/hs/sn_1.sock";
 
-    #ifdef RV_CURL_SOCKET
-    char* socket_path = "/tmp/hs/sn_1.sock";
-
-    if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, socket_path))
-        FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTSET, NULL, "can't set cURL socket path header: %s", curl_err_buf);
-    #endif RV_CURL_SOCKET
+        if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, socket_path))
+            FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTSET, NULL, "can't set cURL socket path header: %s", curl_err_buf);
+    }
 
 #ifdef RV_CURL_DEBUG
     /* Enable cURL debugging output if desired */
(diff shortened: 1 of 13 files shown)


src/rest_vol.c
```diff
@@ -639,6 +647,58 @@ H5_rest_set_connection_information(void)
...
+        FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTALLOC, FAIL, "can't allocate space necessary for placeholder base URL");
+    
+      strncpy(base_URL, URL, URL_len);
+      base_URL[URL_len] = '\0';
+
+      /* Get username/password and authenticate as normal in the env case*/
+      const char *username = getenv("HSDS_USERNAME");
+      const char *password = getenv("HSDS_PASSWORD");
+
+    if (username || password) {
```
jhendersonHDF, line 666: We should probably eventually refactor a bit to share common code between here and below so we aren't duplicating it all.

src/rest_vol.c
```diff
@@ -639,6 +647,58 @@ H5_rest_set_connection_information(void)
...
+    
+      /* This is just a placeholder URL for curl's syntax, its specific value is unimportant */
+      URL = "0";
+      URL_len = 1; 
+
+      if (NULL == (base_URL = (char *) RV_malloc(URL_len + 1)))
+        FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTALLOC, FAIL, "can't allocate space necessary for placeholder base URL");
+    
+      strncpy(base_URL, URL, URL_len);
+      base_URL[URL_len] = '\0';
```
jhendersonHDF, line 660: I'm surprised this works without having a base_URL. I guess when using the domain sockets it's implicitly pointing to the HSDS root domain and then the server can accept URLs in the form of just "/home/test_user1" rather than something such as "https://hsds.hdf.test/home/test_user1/". @jreadey Do you know if this is the case?

test/test_rest_vol.c
```diff
@@ -1493,8 +1493,6 @@ test_unused_file_API_calls(void)
         hid_t                obj_id;
         void                *file_handle;
 
-        if (H5Fis_accessible(filename, fapl_id) >= 0)
```
jhendersonHDF, line 1496: Was there any particular reason for removing this call? Not that it matters since we want to deprecate the old tests anyway, but eventually we'll need to fully support `H5Fis_accessible` in the REST VOL if we don't already.

src/rest_vol_file.c
```diff
@@ -537,12 +539,69 @@ RV_file_specific(void *obj, H5VL_file_specific_args_t *args, hid_t dxpl_id, void
...
+                *ret_is_accessible = TRUE;
+                RV_file_close(ret_file, dxpl_id, NULL);
+            }
+
             break;
 
         /* H5Fdelete */
         case H5VL_FILE_DELETE:
-            FUNC_GOTO_ERROR(H5E_FILE, H5E_UNSUPPORTED, FAIL, "H5Fdelete is unsupported");
+            char  *host_header;
```
jhendersonHDF, line 560: For things that will be malloc'ed, it's a good idea to have these variables exist at function-level scope rather than case-level and then move the free call past the `done:` label so that if an error occurs you can still free these without leaking the memory.

src/rest_vol_file.c
```diff
@@ -537,12 +539,69 @@ RV_file_specific(void *obj, H5VL_file_specific_args_t *args, hid_t dxpl_id, void
 
         /* H5Fis_accessible */
         case H5VL_FILE_IS_ACCESSIBLE:
```
jhendersonHDF, line 541: Whenever you have variables to declare inside a case statement, it's usually a good idea to start a new scope with braces like in the other case statements here. Not doing so can cause issues depending on what you're doing. It looks like the `filename` variable for the `H5VL_FILE_DELETE` case is probably sharing the variable declared in the `H5VL_FILE_IS_ACCESSIBLE` case, but we should add braces to each and separate `filename` variables to make things clearer.

test/test_rest_vol.c
mattjala: If I understand right, this test was expecting `H5Fis_accessible` to always fail, which shouldn't be the case now that it's implemented.

jhendersonHDF: Ah yes, I probably could have figured that out from the test name if I'd looked.. disregard.

src/rest_vol_attr.c
```diff
@@ -1763,6 +1763,19 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
                      */
                     switch (parent_obj_type) {
                         case H5I_FILE:
+                            /* Copy fapl, fcpl, and filepath name to new object */
```
mattjala, line 1766: The reason for this change is that using the same logic for files and groups led to a memory error with `filepath_name`. 

Specifically, the group logic increments the reference count to the `hid_t` resources inside the parent object, such that the next call to close the resource only decreases the reference count. But because a file's `filepath_name` is a raw string, its reference count can't be increased. Thus, the previous logic would double free it.

src/rest_vol_attr.c
```diff
@@ -2019,7 +2035,25 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
...
+            /*
+            RV_object_t *attr_iter_obj = (RV_object_t*) attr_iter_object;
+
+            if (H5Pclose(attr_iter_obj->u.file.fapl_id) < 0)
+                FUNC_DONE_ERROR(H5E_PLIST, H5E_CANTCLOSEOBJ, FAIL, "can't close FAPL");
+            if (H5Pclose(attr_iter_obj->u.file.fcpl_id) < 0)
+                FUNC_DONE_ERROR(H5E_PLIST, H5E_CANTCLOSEOBJ, FAIL, "can't close FCPL");
+
+            if (attr_iter_obj->u.file.filepath_name)
+                RV_free(attr_iter_obj->u.file.filepath_name);
```
jhendersonHDF, line 2049: Was this cleanup code intended to be commented out? Is it handled elsewhere?

mattjala: Yes, the cleanup in this case is handled by H5Fclose

jhendersonHDF (approved):

PR #19, opened 2023-04-03
Fix workflow test execution

mattjala:
This is a squash merge of all the commits from the vol-test-workflow-patch branch to keep the commit history readable

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,15 +2,53 @@ name: Test REST VOL with socket
 
 on: [workflow_dispatch, push]
 
-jobs:                              
+jobs:
   build_and_test_rest_vol:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8"]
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -114,75 +147,49 @@ jobs:
           key: hdf5
           path: ${{github.workspace}}/hdf5install
           
-      - name: Build REST VOL (Autotools)
+      - name: Clone vol-tests repo
+        uses: actions/checkout@v3
+        with:
+          path: ${{github.workspace}}/rest-vol/test/vol-tests/
+          repository: mattjala/vol-tests
```
jhendersonHDF, line 154: Are there still specific fixes in your fork of the VOL tests to make the test run correctly? We should probably base this workflow off of the upstream so that it's useful to see the state of the current VOL tests

jhendersonHDF (approved):

PR #26, opened 2023-04-12
Fix read callback to handle >64kb transfers

mattjala:
curl uses an internal buffer of size 64kb. When the VOL had a payload larger than that to upload, the read data callback would copy the first 64kb of data each time, and never return 0 to indicate the transfer was finished. It would continue to write to the server until the server closed its end of the connection.

Also fixed the read large point selection test to no longer access invalid memory. If run, it now receives a
413 Entity too large error from HSDS instead.

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -1243,9 +1243,8 @@ done:
  *              buffer into cURL's internal buffer when making an HTTP PUT
  *              call to the server.
  *
- * Return:      Amount of bytes equal to the amount given in the upload
- *              info struct on success/0 on failure or if NULL data buffer
- *              is given
+ * Return:      Amount of bytes uploaded, 0 if transfer finished or
+ *              NULL data buffer is given
  *
  * Programmer:  Jordan Henderson
  *              January, 2018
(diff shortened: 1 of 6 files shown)


test/test_rest_vol.c
```diff
@@ -8254,23 +8253,24 @@ test_read_dataset_large_point_selection(void)
         goto error;
     }
 
-    for (i = 0, data_size = 1; i < DATASET_LARGE_READ_TEST_POINT_SELECTION_DSET_SPACE_RANK; i++)
-        data_size *= dims[i];
-    data_size = DATASET_LARGE_READ_TEST_POINT_SELECTION_DSET_DTYPESIZE;
```
jhendersonHDF, line 8259: Looks like all that was needed to fix this was to change this to `data_size *= DATASET_LARGE_READ_TEST_POINT_SELECTION_DSET_DTYPESIZE;`. That said, I've asked @glennsong09 to remove the large I/O tests as we move these tests back into the library since they aren't exactly practical to run.

src/rest_vol.c
```diff
@@ -1257,10 +1256,22 @@ H5_rest_curl_read_data_callback(char *buffer, size_t size, size_t nmemb, void *i
     size_t       max_buf_size = size * nmemb;
     size_t       data_size = 0;
 
+    /* If all bytes sent, indicate transfer is finished */
+    if (uinfo->bytes_sent >= uinfo->buffer_size) {
+        return 0;
+    }
```
jhendersonHDF, line 1262: This block should be moved inside the `if (inptr)` block below so that we don't attempt to dereference `uinfo` in case `inptr` is NULL.

src/rest_vol.c
```diff
@@ -1257,10 +1256,22 @@ H5_rest_curl_read_data_callback(char *buffer, size_t size, size_t nmemb, void *i
...
+
     if (inptr) {
```
jhendersonHDF, line 1264: Instead of checking for buffer overrun each time, I feel like it'd be cleaner and more readable to update this code like:

    if (inptr) {
        size_t bytes_left;

        /* If all bytes sent, indicate transfer is finished */
        if (uinfo->bytes_sent >= uinfo->buffer_size) {
            return 0;
        }

        bytes_left = uinfo->buffer_size - uinfo->bytes_sent;
        data_size = (bytes_left > max_buf_size) ? max_buf_size : bytes_left;

        memcpy(buffer, uinfo->buffer + uinfo->bytes_sent, data_size);

        uinfo->bytes_sent += data_size;
    }

    return data_size;

We could add an assert in the code to catch buffer overruns in debug builds, but otherwise we shouldn't need to check for that case as long as the math adds up.

jhendersonHDF (approved):

Issue #24, opened 2023-04-06
External Links Failing in Vol-Rest


ron-kuhn:
I have a main hdf5 object that has external links to other child hdf5 objects. When using the vol-rest to open a group across the external link, vol-rest recognizes the external link and HSDS actually finds the linked object in the child but the vol-rest sets the main file for the domain in the RV_object_t object instead of the child file. When doing the next operation on the child object, the vol specifies the main file in the URL header. HSDS does find the URI for the child in its cache but returns that the "Object id is not a member of the given domain" and result in failure for the call. 

In don't think this is a hard fix and will fix a couple of my issues.

ron-kuhn:
After some more investigation, inside the RV_group_open function in rest_vol_group.c, the group domain is set to parent's:
`group->domain = parent->domain`
The HTTP response has the name of the child file for external links so it can be parsed by RV_copy_object_URI_callback and compared with the parent and updated in the group.

ron-kuhn:
Made a temporary fix just to test out updating the domain if call to RV_find_object_by_path has to go through an external link. This is only to test if this will fix the problem stated above:

- Add external RV_object_t* external_file = NULL;
- Rename RV_find_object_by_path to RV_find_object_by_path2
- Added code to RV_find_object_by_path2:

`            if (H5L_TYPE_EXTERNAL == link_info.type) {
                /* Unpack the external link's value buffer */
                if (H5Lunpack_elink_val(tmp_link_val, link_val_len, NULL, &ext_filename, &ext_obj_path) < 0)
                    FUNC_GOTO_ERROR(H5E_LINK, H5E_CANTGET, FAIL, "can't unpack external link's value buffer");

                /* Attempt to open the file referenced by the external link using the same access flags as
                 * used to open the file that the link resides within.
                 */
                if (external_file)
                {
                    RV_object_t* external_file_tmp = external_file;
                    external_file = NULL;
                    if (RV_file_close(external_file_tmp, H5P_DEFAULT, NULL) < 0)
                        FUNC_DONE_ERROR(H5E_LINK, H5E_CANTCLOSEOBJ, FAIL, "can't close file referenced by external link");
                }
                
                if (NULL == (external_file = RV_file_open(ext_filename, parent_obj->domain->u.file.intent,
                        parent_obj->domain->u.file.fapl_id, H5P_DEFAULT, NULL)))
[...]

PR #30, opened 2023-04-17
Fix external links for groups, dsets, dtypes

mattjala:
Groups, datasets, and datatypes from an external file now correctly indicate this in their domain field upon being opened. 

H5Oget_info_by_name now correctly retrieves external links' domain.

Fix for #24 , #30

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -1661,22 +1661,16 @@ done:
 herr_t RV_copy_object_URI_and_domain_callback(char *HTTP_response, void *callback_data_in, void *callback_data_out) {
     herr_t ret_value = SUCCEED;
     RV_object_t *parent_object = (RV_object_t*) callback_data_in;
-    void  *URI_domain_ptrs[2];
-    void  *URI_buffer = NULL;
-
-    memcpy(URI_domain_ptrs, callback_data_out, sizeof(void*) * 2);
-    
-    URI_buffer = URI_domain_ptrs[0];
+    loc_info    *loc_info_out = (loc_info*) callback_data_out;
 
     if (parent_object) {
         /* Close reference to previous domain */
-        void **domain_buffer_ptr = (((char*) callback_data_out) + sizeof(void*));
-        RV_file_close(*domain_buffer_ptr, H5P_DEFAULT, NULL);
+        RV_file_close(loc_info_out->domain, H5P_DEFAULT, NULL);
         
-        if (RV_file_create_new_reference(parent_object, domain_buffer_ptr) < 0)
+        if (RV_file_create_new_reference(parent_object, &loc_info_out->domain) < 0)
             FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTCOPY, FAIL, "couldn't copy object domain");
 
-        ret_value = RV_copy_object_URI_callback(HTTP_response, NULL, URI_buffer);
+        ret_value = RV_copy_object_URI_callback(HTTP_response, NULL, loc_info_out->URI);
     } else {
         ret_value = FAIL;
     }
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -1667,8 +1667,8 @@ herr_t RV_copy_object_URI_and_domain_callback(char *HTTP_response, void *callbac
         /* Close reference to previous domain */
         RV_file_close(loc_info_out->domain, H5P_DEFAULT, NULL);
         
-        if (RV_file_create_new_reference(parent_object, &loc_info_out->domain) < 0)
-            FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTCOPY, FAIL, "couldn't copy object domain");
+        loc_info_out->domain = parent_object;
+        parent_object->u.file.ref_count++;
 
         ret_value = RV_copy_object_URI_callback(HTTP_response, NULL, loc_info_out->URI);
     } else {
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -1267,7 +1267,7 @@ H5_rest_curl_read_data_callback(char *buffer, size_t size, size_t nmemb, void *i
         bytes_left = uinfo->buffer_size - uinfo->bytes_sent;
         data_size = (bytes_left > max_buf_size) ? max_buf_size : bytes_left;
 
-        memcpy(buffer, uinfo->buffer + uinfo->bytes_sent, data_size);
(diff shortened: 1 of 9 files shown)


ron-kuhn:
Stinks that all object now will have their own copy of the domain. Too bad this isn't implemented in c++ so that you can use shared_ptr. Copying the domain's RV_object_t  into new memory now pulls in the fapl and fcpl. If these were to be utilized somehow from the children, they have the potential to be corrupt after the the real domain file closes.

ron-kuhn:
Could you add a reference count to the RV_file_t that increment the reference count when copied and RV_file_close will decrement the reference count and close the file resources when reference count is 0. You could use RV_file_close on the domain of all the objects.

ron-kuhn:
@mattjala, nice use of reference counting for the domain. Pull request tested great.  One problem is that MS VS doesn't allow adding to void*. rest-vol.c line 1270 had to be changed to:

`        memcpy(buffer, (const char*)uinfo->buffer + uinfo->bytes_sent, data_size);`

src/rest_vol.c
```diff
@@ -1637,6 +1637,53 @@ RV_parse_response(char *HTTP_response, void *callback_data_in, void *callback_da
...
+ *              and the second being the address of a pointer to a domain buffer of the target object.
+ * 
+ *              callback_data_is provided inside RV_find_object_by_path.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  Matthew Larson
+ *              April, 2023
+ */
+herr_t RV_copy_object_URI_and_domain_callback(char *HTTP_response, void *callback_data_in, void *callback_data_out) {
```
jhendersonHDF, line 1661: It will document the intent of the callback much better to use a struct type containing the two pointers needed here and then pass a pointer to one of those structs in callback_data_out and cast `callback_data_out` to a pointer to that struct type here.

src/rest_vol_file.c
```diff
@@ -653,10 +656,7 @@ RV_file_close(void *file, hid_t dxpl_id, void **req)
     if (H5I_FILE != _file->obj_type)
         FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "not a file");
 
-    if (_file->u.file.filepath_name) {
-        RV_free(_file->u.file.filepath_name);
-        _file->u.file.filepath_name = NULL;
-    }
+    _file->u.file.ref_count--;
 
     if (_file->u.file.fapl_id >= 0) {
```
jhendersonHDF, line 661: If we're doing proper ref. counting, then we probably want to move closing of any file fields like these into the `_file->u.file.ref_count == 0` block as well.

src/rest_vol_attr.c
```diff
@@ -1757,40 +1765,54 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
...
+
+                        if (loc_obj->domain->u.file.fapl_id >= 0) {
+                            if ((loc_obj->domain->u.file.fapl_id != H5P_FILE_ACCESS_DEFAULT) && (H5Iinc_ref(loc_obj->domain->u.file.fapl_id) < 0))
+                                FUNC_DONE_ERROR(H5E_PLIST, H5E_CANTCLOSEOBJ, FAIL, "can't increase ref count of FAPL");
+                        } /* end if */
+
+                        if (loc_obj->domain->u.file.fcpl_id >= 0) {
+                            if ((loc_obj->domain->u.file.fcpl_id != H5P_FILE_CREATE_DEFAULT) && (H5Iinc_ref(loc_obj->domain->u.file.fcpl_id) < 0))
+                                FUNC_DONE_ERROR(H5E_PLIST, H5E_CANTCLOSEOBJ, FAIL, "can't increase ref count of FCPL");
+                        } /* end if */
```
jhendersonHDF, line 1793: The above would need to be replaced by a call to RV_file_create_new_reference, but since I don't think that function is necessary, this can probably just be `loc_obj->domain->u.file.ref_count++;` if the closing of the FAPL and FCPL in file close are moved into the block that checks the ref. count.

src/rest_vol.c
```diff
@@ -2219,9 +2266,14 @@ RV_find_object_by_path(RV_object_t *parent_obj, const char *obj_path,
 #endif
 
         if (ret_value > 0) {
-            if (obj_found_callback && RV_parse_response(response_buffer.buffer,
-                    callback_data_in, callback_data_out, obj_found_callback) < 0)
+            if (obj_found_callback == RV_copy_object_URI_and_domain_callback) {
```
jhendersonHDF, line 2269: Rather than changing this line to look for a specific callback, the code calling this function should be aware that it needs to pass `parent_obj->domain` for `callback_data_in` and those places should be changed instead.

mattjala: In the case of an external link, `parent_obj` will have been reassigned to be an external file at [line 2043](https://github.com/HDFGroup/vol-rest/blob/master/src/rest_vol.c#L2043), so passing `parent_obj->domain` from the code calling this function would always provide the original, potentially incorrect domain.

Changing `callback_data_in` when an external link is found wouldn't be a solution either, as that would break any callbacks which use `callback_data_in` when they're used on an external link.

jhendersonHDF: This seems like an architectural issue. Relying on RV_find_object_by_path to check for a specific callback function and then replace the passed in data with internal data is a fragile solution that will likely lead to bugs if other similar callback functions are added in the future.

mattjala: I was able to resolve this - the server returns the URI and path to the domain for the object, which can be used to determine if an object is external and to construct a new domain object.

src/rest_vol_attr.c
```diff
@@ -1757,40 +1764,54 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
...
```
jhendersonHDF, line 1792: Is there a corresponding close call for the above? Otherwise, it looks like the memory for `loc_obj->domain->u.file` would be leaking and the IDs for `loc_obj->domain->u.file.fapl_id` and `loc_obj->domain->u.file.fcpl_id` would be leaking.

mattjala: The domain is closed inside the close call for each particular type at the end of the function. You're correct these property list IDs were leaking after the change to `RV_file_close` - fixed by removing this incrementing.

jhendersonHDF (approved):

PR #32, opened 2023-04-27
Optimize RV_find_object_by_path

mattjala:
Reduced the number of requests to HSDS made during calls to RV_find_object_by_path.

Most calls now provide the type of object they expect, and use only a single external request.

If the type of the final object is unknown at the time the function is called, then the type is figured out in a single request (previously, this required two calls).

Does not yet handle soft links to other soft links. 

This depends on the changes to HSDS in HDFGroup/HSDS#219 and HDFGroup/HSDS#224

After this change, the icesat2 benchmark completes about 25% faster (5 seconds -> 4 seconds), see HDFGroup/nasa_cloud#4

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -1599,14 +1599,14 @@ done:
     return ret_value;
 } /* end H5_rest_url_encode_path() */
 
-/* Helper function to determine an object's type by its assigned URI */
+/* Helper function to parse an object's type by server response */
 herr_t
 RV_parse_type(char *HTTP_response, void *callback_data_in, void *callback_data_out)
 {
-    yajl_val parse_tree = NULL, key_obj;
-    char    *parsed_object_string;
-    H5I_type_t    *object_type   = (H5I_type_t *) callback_data_out;
-    herr_t   ret_value = SUCCEED;
+    yajl_val    parse_tree = NULL, key_obj;
+    char       *parsed_object_string;
+    H5I_type_t *object_type = (H5I_type_t *)callback_data_out;
+    herr_t      ret_value   = SUCCEED;
 
 #ifdef RV_CONNECTOR_DEBUG
     printf("-> Retrieving object's class from server's HTTP response\n\n");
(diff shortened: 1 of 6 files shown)


mattjala:
Using the `recurse_until_hard` query option added in https://github.com/HDFGroup/hsds/pull/224, `RV_find_object_by_path` can now find its target object in a single request, regardless of whether or not its type is known, or whether the known path is to a symbolic link rather than a hard one. 

To get a metric for how much this reduces overhead on top of the previous change, the icesat2 benchmark time went from ~4 seconds to ~3.7 seconds.

src/rest_vol_datatype.c
```diff
@@ -328,7 +328,7 @@ RV_datatype_open(void *obj, const H5VL_loc_params_t *loc_params, const char *nam
 {
     RV_object_t *parent   = (RV_object_t *)obj;
     RV_object_t *datatype = NULL;
-    H5I_type_t   obj_type = H5I_UNINIT;
```
jhendersonHDF, line 331: Other object open callbacks appear to still be using H5I_UNINIT. Does this work correctly here? Does this need to be changed back or can the other object open callbacks set their expected types ahead of time now?

src/rest_vol.c
```diff
@@ -2307,7 +2360,209 @@ RV_find_object_by_path(RV_object_t *parent_obj, const char *obj_path, H5I_type_t
...
+                                                    callback_data_out, obj_found_callback) < 0) {
+            FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CALLBACK, FAIL, "can't perform callback operation");
+        }
+
+    } /* end if */
+
+    char uri[URI_MAX_LENGTH];
+
+    if (0 > RV_parse_response(response_buffer.buffer, NULL, uri, RV_copy_object_URI_callback))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CANTGET, FAIL, "failed to get URI");
```
jhendersonHDF, line 2542: It looks like this URI object doesn't get used anywhere after being parsed out of the reponse

src/rest_vol.c
```diff
@@ -2307,7 +2360,209 @@ RV_find_object_by_path(RV_object_t *parent_obj, const char *obj_path, H5I_type_t
...
+ *
+ * Return:      Non-negative on success, negative on failure
+ *
+ * Programmer:  Jordan Henderson
+ *              November, 2017
+ */
+htri_t
+RV_find_object_by_path2(RV_object_t *parent_obj, const char *obj_path, H5I_type_t *target_object_type,
+                        herr_t (*obj_found_callback)(char *, void *, void *), void *callback_data_in,
+                        void *callback_data_out)
```
jhendersonHDF, line 2394: There's a lot of duplication between the two versions of this function that probably could have been handled by just checking the server version and determining the correct URL to use and whether or not to make use of the RV_parse_type callback, but we can worry about cleaning this up later.

src/rest_vol_datatype.c
mattjala: `RV_find_object_by_path2` stores the type of the object it finds, but no longer needs the initial value when making the request.

`RV_find_object_by_path1` should still need H5I_UNINIT in some cases - it looks like `RV_datatype_open` isn't used by any of the active tests right now. I'll set this back to H5I_UNINIT for consistency with any version of HSDS.

jhendersonHDF (approved):

PR #33, opened 2023-05-04
Store GCPL information in server

mattjala:
Stores GCPL information given by user in the server, and retrieves it when the group is opened. 

Depends on the server returning the `creationProperties` field for groups, which was added in HDFGroup/hsds#220


src/rest_vol.c
```diff
@@ -2404,6 +2416,61 @@ RV_find_object_by_path(RV_object_t *parent_obj, const char *obj_path, H5I_type_t
...
+    if (NULL == (parsed_string = YAJL_GET_STRING(key_obj)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "creationProperties was NULL");
+
+    if (NULL == (*buf_out = RV_malloc(strlen(parsed_string) + 1)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CANTALLOC, FAIL, "failed to allocate memory for creationProperties");
+
+    if (NULL == (memcpy(*buf_out, parsed_string, strlen(parsed_string) + 1)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_SYSERRSTR, FAIL, "failed to copy creationProperties");
+
+done:
```
jhendersonHDF, line 2467: memcpy won't fail and so won't cause an issue here, but it would be good to check for failure so that `*buf_out` can be freed and set to NULL if something fails past allocating it.

src/rest_vol_group.c
```diff
@@ -349,6 +378,25 @@ RV_group_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name,
     printf("-> Found group by given path\n\n");
 #endif
 
+    obj_type = H5I_UNINIT;
+
+    search_ret = RV_find_object_by_path(parent, name, &obj_type, RV_copy_creation_properties_callback, NULL,
+                                        &base64_binary_gcpl);
```
jhendersonHDF, line 384: You might want to consider having a combined `RV_copy_object_URI_and_domain_callback` and `RV_copy_creation_properties_callback` in the future so we aren't making multiple calls to RV_find_object_by_path for the same object.

src/rest_vol_group.c
```diff
@@ -349,6 +378,25 @@ RV_group_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name,
...
+    if (!search_ret || search_ret < 0)
+        FUNC_GOTO_ERROR(H5E_SYM, H5E_PATH, NULL, "can't locate group by path");
+
+    /* Decode creation properties */
+    if (RV_base64_decode(base64_binary_gcpl, strlen(base64_binary_gcpl), &binary_gcpl, &binary_gcpl_size) < 0)
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CANTDECODE, NULL, "can't decode gcpl from base64");
+
+    /* Set up a GCPL for the group so that H5Gget_create_plist() will function correctly */
+    /* XXX: Because the HDF5 REST API does not support storing creationProperties for groups,
```
jhendersonHDF, line 393: Looks like the TODO parts of this comment can be removed now

src/rest_vol_group.c
mattjala: I'm a little wary of this - we already have distinct `RV_copy_object_URI_callback` and `RV_copy_object_URI_and_domain_callback` functions, so `RV_copy_URI_domain_creation_properties` would bring us up to three functions capturing a different combination of data, and it would only be used in this one place. Would one function that tries to get a wide selection of fields, but fails gracefully for ones that aren't defined be a decent solution?

src/rest_vol.c
jhendersonHDF: In general only test code uses the `error:` label. In library code the `done:` label is used as cleanup and so checking for `ret_value < 0` here would be the main way to tell if something failed.

src/rest_vol_group.c
jhendersonHDF: I agree, but it's a product of the architecture needing to change as functionality is added. It might make the client side code more complex, but I'd say that's probably dwarfed by the need to make at least one extra network request in this case or possibly more if an optimized RV_find_object_by_path can't be used for whatever reason.

Experimenting with a single function to grab everything needed is reasonable. You could also have a composition of callback functions as was done for the `RV_copy_object_URI_and_domain_callback` callback, but the issue with that is you end up with several layers recreating parse trees for the same response buffer just to grab a single field out of it.

src/rest_vol.c
```diff
@@ -2404,6 +2302,186 @@ RV_find_object_by_path(RV_object_t *parent_obj, const char *obj_path, H5I_type_t
...
+    if (NULL == (*GCPL_buf = RV_malloc(strlen(parsed_string) + 1)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CANTALLOC, FAIL, "failed to allocate memory for creationProperties");
+
+    if (NULL == (memcpy(*GCPL_buf, parsed_string, strlen(parsed_string) + 1)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_SYSERRSTR, FAIL, "failed to copy creationProperties");
+
+done:
+
+    if (ret_value < 0) {
+        free(*GCPL_buf);
```
jhendersonHDF, line 2351: I don't think this is important enough to hold this PR, but note that this function should probably be updated to allocate the GCPL buf and assign it to a local variable that was first initialized to NULL, then make the free call here on the local variable rather than the `GCPL_buf` pointer passed in. Otherwise, a failure in this function before the call to `RV_malloc` will cause a segfault or double free error if the caller didn't initialize `*GCPL_buf` to NULL. Typically, to prevent disturbing the caller's state on failure, I'll wait to assign the local pointer to the returned one until the last statement in the body of the function.

jhendersonHDF (approved):

PR #35, opened 2023-05-12
Keep track of HSDS version

mattjala:
This will be useful for reverting to old `RV_find_object_by_path` behavior for old versions of HSDS that don't include the changes from HDFGroup/HSDS#224

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -84,6 +84,10 @@ size_t H5_rest_curr_alloc_bytes;
  */
 struct response_buffer response_buffer;
 
+/* Structure to keep track of server version that REST VOL is communicating with.
+ * Updated each time a file is updated or opened. */
+server_api_version server_version;
+
 /* Authentication information for authenticating
  * with Active Directory.
  */
@@ -126,6 +130,9 @@ const char *domain_keys[] = {"domain", (const char *)0};
  * of RV_base64_encode supplies a 0-sized buffer.
  */
 #define BASE64_ENCODE_DEFAULT_BUFFER_SIZE 33554432 /* 32MB */
+
+/* JSON key to retrieve the version of server from a request to a file. */
+const char *server_version_keys[] = {"version", (const char *)0};
 
 /* Internal initialization/termination functions which are called by
  * the public functions H5rest_init() and H5rest_term() */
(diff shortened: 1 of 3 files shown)


src/rest_vol.h
```diff
@@ -373,6 +376,17 @@ struct response_buffer {
 };
 extern struct response_buffer response_buffer;
 
+/* Structure that keeps track of semantic version. */
+typedef struct version {
+    size_t major;
+    size_t minor;
+    size_t patch;
+} version;
```
jhendersonHDF, line 384: I'd probably call this structure something less generic like `server_api_version` so it doesn't get confused among other possible "versions" and this struct should be a part of the `RV_file_t` structure rather than being an extern variable. Otherwise, opening multiple files from different server versions will likely lead to problems.

src/rest_vol.c
```diff
@@ -119,6 +123,9 @@ const char *dataspace_max_dims_keys[] = {"shape", "maxdims", (const char *)0};
 /* JSON keys to retrieve the path of a domain */
 const char *domain_keys[] = {"domain", (const char *)0};
 
+/* JSON key to retrieve the version of HSDS from a request to a file. */
+extern const char *hsds_version_keys[] = {"version", (const char *)0};
```
jhendersonHDF, line 127: no extern needed here

src/rest_vol.c
```diff
@@ -2761,6 +2768,68 @@ RV_convert_dataspace_shape_to_JSON(hid_t space_id, char **shape_body, char **max
...
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "failed to parse HSDS version");
+
+    if (!YAJL_IS_STRING(key_obj))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "parsed HSDS version is not a string");
+
+    if (NULL == (version_response = YAJL_GET_STRING(key_obj)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "HSDS version was NULL");
+
+    /* Parse HSDS version into struct */
+    if (NULL == (version_field = strtok_r(version_response, ".", &version_response)))
```
jhendersonHDF, line 2803: Using strtok_r is reasonable here for the future, but you should use a separate variable for its third argument for storing state rather than reusing a pointer to the same pointer.

Issue #34, opened 2023-05-08
Link/Attribute Access by Creation Order


mattjala:
The REST VOL should support accessing links/attribute by creation order, instead of by lexical order on their keys. HSDS stores the creation order parameter as of HDFGroup/HSDS#225, so this can be implemented.

mattjala:
Implemented in #37

PR #37, opened 2023-05-30
Implement 'by index' functions

mattjala:
Implemented support for H5Gget_info_by_name, H5Oget_info_by_idx, H5Aget_info_by_idx,
H5Aget_name_by_idx, and H5Aopen_by_idx (#34)

Updated test_rest_vol.c to stop expecting some of those functions to fail.

Patched a memory error in H5Gget_info_by_name, and
cleaned up some things that were causing warnings on build.


src/rest_vol.c
```diff
@@ -2597,17 +2602,21 @@ RV_parse_creation_properties_callback(yajl_val parse_tree, char **GCPL_buf)
...
 
-    if (NULL == (*GCPL_buf = RV_malloc(strlen(parsed_string) + 1)))
+    if (NULL == (GCPL_buf_local = RV_malloc(strlen(parsed_string) + 1)))
         FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CANTALLOC, FAIL, "failed to allocate memory for creationProperties");
 
-    if (NULL == (memcpy(*GCPL_buf, parsed_string, strlen(parsed_string) + 1)))
+    if (NULL == (memcpy(GCPL_buf_local, parsed_string, strlen(parsed_string) + 1)))
         FUNC_GOTO_ERROR(H5E_OBJECT, H5E_SYSERRSTR, FAIL, "failed to copy creationProperties");
 
+    if (NULL == (*GCPL_buf_out = GCPL_buf_local))
```
jhendersonHDF, line 2611: No need to check for NULL here, just assign `*GCPL_buf_out` to `GCPL_buf_local`. Though, this function should probably have a check that `GCPL_buf_out` isn't NULL to prevent a possible segfault.

src/rest_vol.c
```diff
@@ -2740,6 +2749,163 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+    switch (idx_params->order) {
+        case (H5_ITER_DEC):
+            if (NULL == (link_obj = key_obj->u.array.values[key_obj->u.object.len - 1 - index]))
+                FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "selected link was NULL");
+            break;
+
+        case (H5_ITER_NATIVE):
+        case (H5_ITER_INC):
+        case (H5_ITER_N):
+        case (H5_ITER_UNKNOWN):
```
jhendersonHDF, line 2799: cases `H5_ITER_N`, `H5_ITER_UNKNOWN` and default should return an error. Only cases `H5_ITER_NATIVE` and `H5_ITER_INC` should have the below implementation.

src/rest_vol.c
```diff
@@ -2740,6 +2749,163 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+                FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "failed to get link name");
+    }
+
+    if (NULL == parsed_link_name)
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "server response didn't contain link name");
+
+    if (NULL == (*link_name = RV_malloc(strlen(parsed_link_name) + 1)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "failed to allocate memory for link name");
+
+    if (NULL == (memcpy(*link_name, parsed_link_name, strlen(parsed_link_name) + 1)))
```
jhendersonHDF, line 2822: No need to check for NULL here, just do the memcpy. But this function should also have a check that `*link_name` isn't NULL.

src/rest_vol.c
```diff
@@ -2740,6 +2749,163 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+ *              in the server's response by index. It allocates heap memory for the name,
+ *              and returns a pointer to the memory by callback_data_out.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  Matthew Larson
+ *              May, 2023
+ */
+herr_t
+RV_copy_attribute_name_by_index(char *HTTP_response, void *callback_data_in, void *callback_data_out)
```
jhendersonHDF, line 2850: The same comments from `RV_copy_link_name_by_index` seem to apply for this function as well.

src/rest_vol_group.c
```diff
@@ -542,6 +542,9 @@ RV_group_get(void *obj, H5VL_group_get_args_t *args, hid_t dxpl_id, void **req)
                         FUNC_GOTO_ERROR(H5E_SYM, H5E_SYSERRSTR, FAIL,
                                         "H5Gget_info_by_name request URL size exceeded maximum URL size");
 
+                    if (loc_info_out.GCPL_base64)
```
jhendersonHDF, line 545: In this function, `loc_info_out` should be initialized to 0 with `loc_info_out = {0};` and then this free call should be duplicated in the `done:` section so we don't miss cleaning up memory on failure. In that case, this free call here should set `loc_info_out.GCPL_base64` to NULL after freeing it so we don't accidentally try to double-free the pointer.

src/rest_vol_link.c
```diff
@@ -609,10 +608,17 @@ RV_link_get(void *obj, const H5VL_loc_params_t *loc_params, H5VL_link_get_args_t
...
             by_idx_data.iter_function.link_iter_op = NULL;
             by_idx_data.op_data                    = NULL;
 
+            /* Avoid compiler warnings about const pointer reassignment */
+            hsize_t idx_p;
+
+            if (0 > memcpy(&idx_p, &loc_params->loc_data.loc_by_idx.n, sizeof(hsize_t)))
+                FUNC_GOTO_ERROR(H5E_LINK, H5E_SYSERRSTR, FAIL, "failed to copy loc by idx info");
+
+            by_idx_data.idx_p = &idx_p;
```
jhendersonHDF, line 620: This warning should be solved in a different way. Here, we're returning the address of a local variable which is going to segfault at some point since the variable won't exist any more once we exit this function.

src/rest_vol_link.c
```diff
@@ -609,10 +608,17 @@ RV_link_get(void *obj, const H5VL_loc_params_t *loc_params, H5VL_link_get_args_t
...
             by_idx_data.index_type                 = loc_params->loc_data.loc_by_idx.idx_type;
             by_idx_data.iter_order                 = loc_params->loc_data.loc_by_idx.order;
-            by_idx_data.idx_p                      = &loc_params->loc_data.loc_by_idx.n;
             by_idx_data.iter_function.link_iter_op = NULL;
             by_idx_data.op_data                    = NULL;
 
+            /* Avoid compiler warnings about const pointer reassignment */
+            hsize_t idx_p;
+
+            if (0 > memcpy(&idx_p, &loc_params->loc_data.loc_by_idx.n, sizeof(hsize_t)))
```
jhendersonHDF, line 617: No need to check anything here, just do the `memcpy`. `memcpy` returns a pointer to `&idx_p` so checking for `< 0` here isn't correct. We likely don't need to do any `memcpy` here to fix this warning though.

test/test_rest_vol.c
```diff
@@ -2493,16 +2493,16 @@ test_create_attribute_on_root(void)
 #endif
 
         if ((/*attr_id = */ H5Aopen_by_idx(file_id, "/", H5_INDEX_NAME, H5_ITER_INC, 0, H5P_DEFAULT,
```
jhendersonHDF, line 2495: The code above this point that surrounds calls to `H5Aclose` with `#if 0` should be re-enabled so that the IDs for the attributes from the previous tests get closed correctly. Then, all these lines where `attr_id` and `attr_id2` are commented out should be re-enabled so that the attribute IDs are properly assigned and later closed. Otherwise, we're leaking several IDs from the library. There are at least two instances of this in this file that I can see.

src/rest_vol_attr.c
```diff
@@ -479,7 +480,130 @@ RV_attr_open(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_na
...
+            if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_HTTPGET, 1))
+                FUNC_GOTO_ERROR(H5E_ATTR, H5E_CANTSET, NULL, "can't set up cURL to make HTTP GET request: %s",
+                                curl_err_buf);
+            if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_URL, request_url))
+                FUNC_GOTO_ERROR(H5E_ATTR, H5E_CANTSET, NULL, "can't set cURL request URL: %s", curl_err_buf);
+
+            CURL_PERFORM(curl, H5E_ATTR, H5E_CANTGET, NULL);
+
+            /* Avoid compiler warnings about const pointer reassignment */
+            if (0 > memcpy(&idx_args, &loc_params->loc_data.loc_by_idx, sizeof(H5VL_loc_by_idx_t)))
```
jhendersonHDF, line 590: Same issue with the `memcpy` check here

src/rest_vol_attr.c
```diff
@@ -479,7 +480,130 @@ RV_attr_open(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_na
...
+                FUNC_GOTO_ERROR(H5E_LINK, H5E_SYSERRSTR, NULL, "failed to copy loc by idx info");
+
+            if (0 > RV_parse_response(response_buffer.buffer, (void *)&idx_args, &found_attr_name,
+                                      RV_copy_attribute_name_by_index))
+                FUNC_GOTO_ERROR(H5E_ATTR, H5E_PARSEERROR, NULL, "failed to retrieve attribute names");
+
+            if (host_header)
+                RV_free(host_header);
+            if (url_encoded_attr_name)
+                curl_free(url_encoded_attr_name);
```
jhendersonHDF, line 600: These two variables should be set to NULL after freeing so we don't accidentally double-free them

test/test_rest_vol.c
```diff
@@ -2963,28 +2957,26 @@ test_create_attribute_on_datatype(void)
...
-#endif
 
 #ifdef RV_CONNECTOR_DEBUG
         puts("Attempting to open the attributes with H5Aopen_by_idx\n");
 #endif
 
-        if ((/*attr_id = */ H5Aopen_by_idx(type_id, "/" ATTRIBUTE_CREATE_ON_DATATYPE_DTYPE_NAME,
-                                           H5_INDEX_NAME, H5_ITER_INC, 0, H5P_DEFAULT, H5P_DEFAULT)) >= 0) {
+        if ((attr_id =
+                 H5Aopen_by_idx(type_id, ".", H5_INDEX_NAME, H5_ITER_INC, 0, H5P_DEFAULT, H5P_DEFAULT)) < 0) {
```
jhendersonHDF, line 2970: These tests were intentionally using an absolute path to try and catch any issues there, but as a general note, you should try this PR with the API tests in the library or the external VOL tests repo since they do a better job of testing the by_index functions.

src/rest_vol_attr.c
```diff
@@ -479,7 +487,111 @@ RV_attr_open(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_na
...
+                case (H5_INDEX_UNKNOWN):
+                default:
+                    FUNC_GOTO_ERROR(H5E_ATTR, H5E_CANTALLOC, NULL, "unsupported index type specified");
+                    break;
+            }
+
+            /* Make additional request to server to determine attribute name by index */
+            if (!strcmp(loc_params->loc_data.loc_by_idx.name, ".")) {
+                attribute->u.attribute.parent_obj_type = parent->obj_type;
+                strncpy(attribute->u.attribute.parent_obj_URI, parent->URI, URI_MAX_LENGTH);
```
jhendersonHDF, line 519: Do we need to keep a separate copy of the parent object's URI rather than just retrieving it on demand?

src/rest_vol_object.c
```diff
@@ -726,3 +1261,808 @@ RV_get_object_info_callback(char *HTTP_response, void *callback_data_in, void *c
...
+    const object_table_entry *_object1 = (const object_table_entry *)object1;
+    const object_table_entry *_object2 = (const object_table_entry *)object2;
+
+    return ((_object1->crt_time > _object2->crt_time) - (_object1->crt_time < _object2->crt_time));
+} /* end H5_rest_cmp_objects_by_creation_order_inc() */
+
+void
+RV_free_visited_object_hash_table_key(rv_hash_table_key_t value)
+{
+}
```
jhendersonHDF, line 1292: If this function isn't necessary for this file, it should just be deleted.

src/rest_vol.c
```diff
@@ -1918,35 +1931,41 @@ RV_copy_object_URI_callback(char *HTTP_response, void *callback_data_in, void *c
...
+    char              *host_header           = NULL;
+    char              *path_dirname          = NULL;
+    char              *tmp_link_val          = NULL;
+    char              *url_encoded_path_name = NULL;
+    char              *ext_filename          = NULL;
+    char              *ext_obj_path          = NULL;
+    char               request_url[URL_MAX_LENGTH];
+    long               http_response;
+    int                url_len = 0;
+    server_api_version version = parent_obj->domain->u.file.server_version;
```
jhendersonHDF, line 1966: This should be moved down so that we don't try to de-reference `parent_obj` before we've checked that it's non-NULL.

src/rest_vol.c
```diff
@@ -2597,17 +2332,19 @@ RV_parse_creation_properties_callback(yajl_val parse_tree, char **GCPL_buf)
     if (NULL == (parsed_string = YAJL_GET_STRING(key_obj)))
         FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "creationProperties was NULL");
 
-    if (NULL == (*GCPL_buf = RV_malloc(strlen(parsed_string) + 1)))
+    if (NULL == (GCPL_buf_local = RV_malloc(strlen(parsed_string) + 1)))
         FUNC_GOTO_ERROR(H5E_OBJECT, H5E_CANTALLOC, FAIL, "failed to allocate memory for creationProperties");
 
-    if (NULL == (memcpy(*GCPL_buf, parsed_string, strlen(parsed_string) + 1)))
+    if (NULL == (memcpy(GCPL_buf_local, parsed_string, strlen(parsed_string) + 1)))
```
jhendersonHDF, line 2338: No need to check for NULL here

src/rest_vol.c
```diff
@@ -2597,17 +2332,19 @@ RV_parse_creation_properties_callback(yajl_val parse_tree, char **GCPL_buf)
...
         FUNC_GOTO_ERROR(H5E_OBJECT, H5E_SYSERRSTR, FAIL, "failed to copy creationProperties");
 
+    *GCPL_buf_out = GCPL_buf_local;
+
 done:
 
     if (ret_value < 0) {
-        free(*GCPL_buf);
-        *GCPL_buf = NULL;
+        free(GCPL_buf_local);
```
jhendersonHDF, line 2346: RV_free rather than free. Though we should revisit whether the wrapper functions need to be kept around

src/rest_vol.c
```diff
@@ -460,6 +475,13 @@ H5_rest_init(hid_t vipl_id)
     if ((H5_rest_attr_table_iter_err_min_g =
              H5Ecreate_msg(H5_rest_err_class_g, H5E_MINOR, "Can't iterate through attribute table")) < 0)
         FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTINIT, FAIL, "can't create message for attribute iteration error");
+    if ((H5_rest_object_table_err_min_g =
+             H5Ecreate_msg(H5_rest_err_class_g, H5E_MINOR, "Can't build table of objects for iteration")) < 0)
+        FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTINIT, FAIL,
+                        "can't create error message for object table build error");
+    if ((H5_rest_object_table_iter_err_min_g =
+             H5Ecreate_msg(H5_rest_err_class_g, H5E_MINOR, "Can't iterate through object table")) < 0)
+        FUNC_GOTO_ERROR(H5E_VOL, H5E_CANTINIT, FAIL, "can't create message for object iteration error");
```
jhendersonHDF, line 484: These IDs need to be closed in H5rest_term like the other error message IDs.

src/rest_vol.c
```diff
@@ -2720,7 +2457,15 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
-        RV_file_close(loc_info_out->domain, H5P_DEFAULT, NULL);
+        /* Assume that original domain and external domain have the same server version.
+         * This will always be true unless it becomes possible for external links to point to
+         * objects on different servers entirely. */
+        memcpy(&new_domain->u.file.server_version, &loc_info_out->domain->u.file.server_version,
+               sizeof(server_api_version));
+
+        if (RV_file_close(loc_info_out->domain, H5P_DEFAULT, NULL) < 0)
+            FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTCLOSEFILE, FAIL,
+                            "failed to allocate memory for new domain path");
```
jhendersonHDF, line 2468: wrong error message here

src/rest_vol.c
```diff
@@ -2732,14 +2477,258 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+{
+    yajl_val           parse_tree = NULL, key_obj = NULL, link_obj = NULL;
+    const char        *parsed_link_name   = NULL;
+    char              *parsed_link_buffer = NULL;
+    H5VL_loc_by_idx_t *idx_params         = (H5VL_loc_by_idx_t *)callback_data_in;
+    hsize_t            index              = idx_params->n;
+    char             **link_name          = (char **)callback_data_out;
+    const char        *curr_key           = NULL;
+    herr_t             ret_value          = SUCCEED;
+
```
jhendersonHDF, line 2511: `idx_params` should also be checked as non-NULL here and `index` shouldn't be assigned until it's been verified that `idx_params` is non-NULL.

src/rest_vol.c
```diff
@@ -2732,14 +2477,258 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+    if (NULL == parsed_link_name)
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "server response didn't contain link name");
+
+    if (NULL == (parsed_link_buffer = RV_malloc(strlen(parsed_link_name) + 1)))
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "failed to allocate memory for link name");
+
+    memcpy(parsed_link_buffer, parsed_link_name, strlen(parsed_link_name) + 1);
+
+    *link_name = parsed_link_buffer;
```
jhendersonHDF, line 2566: Same comment as in other functions where a local buffer should be used and assigned to `*link_name` and the local buffer pointer should be freed on failure, not the pointer passed in.

src/rest_vol.c
```diff
@@ -2732,14 +2477,258 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+RV_copy_link_URI_by_index(char *HTTP_response, void *callback_data_in, void *callback_data_out)
+{
+    yajl_val           parse_tree = NULL, key_obj = NULL, link_obj = NULL;
+    const char        *parsed_link_URI = NULL;
+    H5VL_loc_by_idx_t *idx_params      = (H5VL_loc_by_idx_t *)callback_data_in;
+    hsize_t            index           = idx_params->n;
+    char             **link_URI        = (char **)callback_data_out;
+    const char        *curr_key        = NULL;
+    herr_t             ret_value       = SUCCEED;
+
```
jhendersonHDF, line 2601: Same comment about checking `idx_params` as non-NULL here and assigning `index` after we're sure `idx_params` is non-NULL

src/rest_vol.c
```diff
@@ -2732,14 +2477,258 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+    switch (idx_params->order) {
+        case (H5_ITER_DEC):
+            if (NULL == (link_obj = key_obj->u.array.values[key_obj->u.object.len - 1 - index]))
+                FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "selected link was NULL");
+            break;
+
+        case (H5_ITER_NATIVE):
+        case (H5_ITER_INC):
+        case (H5_ITER_N):
```
jhendersonHDF, line 2625: Same comment as above where the `H5_ITER_N` and `H5_ITER_UNKNOWN` and `default` cases should return an error and only `H5_ITER_NATIVE` and `H5_ITER_INC` should have the below implementation.

src/rest_vol.c
```diff
@@ -2732,14 +2477,258 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+
+        if (!strcmp(curr_key, "id"))
+            if (NULL == (parsed_link_URI = YAJL_GET_STRING(link_obj->u.object.values[i])))
+                FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "failed to get link URI");
+    }
+
+    if (NULL == parsed_link_URI)
+        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_PARSEERROR, FAIL, "server response didn't contain link URI");
+
+    if (NULL == (memcpy(*link_URI, parsed_link_URI, strlen(parsed_link_URI) + 1)))
```
jhendersonHDF, line 2646: No need to check for NULL here

jhendersonHDF, line 2646: Since the buffer being copied into wasn't allocated by this function, it can be dangerous to copy into it based on `strlen(parsed_link_URI) + 1` since we don't necessarily know how big `*link_URI` might be.

src/rest_vol.c
```diff
@@ -3396,6 +3385,99 @@ RV_parse_server_version(char *HTTP_response, void *callback_data_in, void *callb
...
+ *              to return the string. The given pointer should not point at
+ *              any allocated memory, as its old value is overwritten.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  Matthew Larson
+ *              May, 2023
+ */
+herr_t
+RV_set_object_type_header(H5I_type_t parent_obj_type, char **parent_obj_type_header)
```
jhendersonHDF, line 3444: This should be declared with `const char **parent_obj_type_header`

test/test_rest_vol.c
mattjala: I've been using the external repo for testing, this is just to keep the CI tests in github happy.

src/rest_vol_attr.c
mattjala: `attribute->u.attribute.parent_obj_URI` is a field on the `RV_attr_t` struct, so that the object an attribute is attached to can be found from the attribute itself at any point (e.g. any time after `RV_attr_open` returns the attribute)

src/rest_vol.c
```diff
@@ -1470,14 +1498,21 @@ H5_rest_url_encode_path(const char *path)
     char     *url_encoded_path_component = NULL;
     char     *token;
     char     *cur_pos;
-    char     *tmp_buffer = NULL;
-    char     *ret_value  = NULL;
+    char     *path = NULL;
+    _path;
```
jhendersonHDF, line 1502: Looks like this empty statement snuck in

src/rest_vol.c
```diff
@@ -1470,14 +1498,21 @@ H5_rest_url_encode_path(const char *path)
...
+    char *ret_value  = NULL;
 
-    if (!path)
+    if (!_path)
         FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, NULL, "path was NULL");
 
+    if ((path = RV_malloc(strlen(_path) + 1)) == NULL)
+        FUNC_GOTO_ERROR(H5E_PATH, H5E_CANTALLOC, NULL, "can't allocate memory for path copy");
+
+    strncpy(path, _path, strlen(_path) + 1);
```
jhendersonHDF, line 1512: This is one way to fix the warning, but allocating a separate buffer and copying the string into it isn't necessary to fix what I'm assuming is a warning about const here

src/rest_vol.c
```diff
@@ -2732,14 +2490,191 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+
+    memcpy(parsed_string_buffer, parsed_string, strlen(parsed_string) + 1);
+
+    *attr_name = parsed_string_buffer;
+done:
+    if (parse_tree)
+        yajl_tree_free(parse_tree);
+
+    if (ret_value < 0) {
+        RV_free(*attr_name);
```
jhendersonHDF, line 2671: `RV_free(*attr_name);` -> `RV_free(parsed_string_buffer);`

src/rest_vol.c
```diff
@@ -2732,14 +2490,191 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
...
+ *              May, 2023
+ */
+herr_t
+RV_copy_attribute_name_by_index(char *HTTP_response, void *callback_data_in, void *callback_data_out)
+{
+    yajl_val           parse_tree           = NULL, key_obj;
+    const char        *parsed_string        = NULL;
+    char              *parsed_string_buffer = NULL;
+    H5VL_loc_by_idx_t *idx_params           = (H5VL_loc_by_idx_t *)callback_data_in;
+    hsize_t            index                = idx_params->n;
```
jhendersonHDF, line 2616: should check for non-NULL `idx_params` before assigning `index`

src/rest_vol_attr.c
```diff
@@ -117,6 +116,10 @@ RV_attr_create(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_
 
     /* If this is a call to H5Acreate_by_name, locate the real parent object */
     if (H5VL_OBJECT_BY_NAME == loc_params->type) {
+        // TODO: Proper way to verify a plist?
+        if (H5I_INVALID_HID == loc_params->loc_data.loc_by_name.lapl_id)
```
jhendersonHDF, line 120: The proper way would probably be either to use H5Iobject_verify or H5Iis_valid. But since H5Iobject_verify can't currently be called on library types and H5Iis_valid may not always return the right result, this is probably fine for now.

src/rest_vol_attr.c
```diff
@@ -1271,7 +1375,115 @@ RV_attr_get(void *obj, H5VL_attr_get_args_t *args, hid_t dxpl_id, void **req)
...
+                    CURL_PERFORM(curl, H5E_ATTR, H5E_CANTGET, FAIL);
+
+                    if (0 > RV_parse_response(response_buffer.buffer, &loc_params->loc_data.loc_by_idx,
+                                              &found_attr_name, RV_copy_attribute_name_by_index))
+                        FUNC_GOTO_ERROR(H5E_ATTR, H5E_PARSEERROR, FAIL, "failed to retrieve attribute names");
+
+                    if (host_header)
+                        RV_free(host_header);
+                    if (url_encoded_attr_name)
+                        curl_free(url_encoded_attr_name);
```
jhendersonHDF, line 1472: `host_header` and `url_encoded_attr_name` should be set to NULL so they aren't double-freed

src/rest_vol_attr.c
```diff
@@ -1351,7 +1563,118 @@ RV_attr_get(void *obj, H5VL_attr_get_args_t *args, hid_t dxpl_id, void **req)
...
+                        strncpy(name_buf, found_attr_name, name_buf_size - 1);
+                        name_buf[name_buf_size - 1] = '\0';
+                    } /* end if */
+
+                    if (host_header)
+                        RV_free(host_header);
+                    host_header = NULL;
+
+                    if (url_encoded_attr_name)
+                        curl_free(url_encoded_attr_name);
```
jhendersonHDF, line 1671: `url_encoded_attr_name` should be set to NULL so it isn't double-freed on error

src/rest_vol_link.c
```diff
@@ -215,6 +217,14 @@ RV_link_create(H5VL_link_create_args_t *args, void *obj, const H5VL_loc_params_t
 
         /* H5Lcreate_soft */
         case H5VL_LINK_CREATE_SOFT: {
+            if (loc_params->type == H5VL_OBJECT_BY_NAME &&
+                H5I_INVALID_HID == loc_params->loc_data.loc_by_name.lapl_id)
+                FUNC_GOTO_ERROR(H5E_ATTR, H5E_BADVALUE, FAIL, "invalid LAPL");
+
+            if (loc_params->type == H5VL_OBJECT_BY_IDX &&
+                H5I_INVALID_HID == loc_params->loc_data.loc_by_idx.lapl_id)
+                FUNC_GOTO_ERROR(H5E_ATTR, H5E_BADVALUE, FAIL, "invalid LAPL");
```
jhendersonHDF, line 226: Looks like these are just duplicating the checks from previously

src/rest_vol_object.c
```diff
@@ -526,8 +678,13 @@ RV_object_get(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_get_ar
...
 
-    RV_file_close(loc_info.domain, H5P_DEFAULT, NULL);
-    free(loc_info.GCPL_base64);
+    if (found_object_name) {
+        RV_free(found_object_name);
+        found_object_name = NULL;
+    }
+
+    RV_file_close(loc_info_out.domain, H5P_DEFAULT, NULL);
+    free(loc_info_out.GCPL_base64);
```
jhendersonHDF, line 687: Looks like RV_free is used elsewhere. Not sure which is the correct one here

src/rest_vol_object.c
```diff
@@ -578,10 +746,368 @@ RV_object_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_s
             break;
 
         /* H5Ovisit(_by_name) */
-        case H5VL_OBJECT_VISIT:
-            FUNC_GOTO_ERROR(H5E_OBJECT, H5E_UNSUPPORTED, FAIL,
-                            "H5Ovisit and H5Ovisit_by_name are unsupported");
+        case H5VL_OBJECT_VISIT: {
+            const char *parent_object_header = NULL;
+            iter_data   object_iter_data;
+            hsize_t     idx_p = 0;
```
jhendersonHDF, line 752: This variable can just be eliminated since there's no index pointer saving capability for H5Ovisit

src/rest_vol_object.c
```diff
@@ -578,10 +746,368 @@ RV_object_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_s
...
+                case (H5VL_OBJECT_BY_SELF): {
+
+                    /* Since the VOL doesn't directly pass down the group's hid_t, explicitly open the group
+                     * here so that a valid hid_t can be passed to the user's link iteration callback.
+                     */
+
+                    switch (loc_obj->obj_type) {
+                        case H5I_FILE:
+                        case H5I_GROUP:
+                            if (NULL == (iter_object = RV_group_open(loc_obj, loc_params, ".", H5P_DEFAULT,
```
jhendersonHDF, line 775: This approach is fine for now since it's what the link code does as well, but eventually it should be switched to use H5VLwrap_register. You can look at the DAOS VOL as an example.

src/rest_vol_object.c
```diff
@@ -578,10 +746,368 @@ RV_object_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_s
...
+                            parent_object_header = "datatypes";
+                            iter_object_type     = H5I_DATATYPE;
+                            break;
+
+                        case H5I_ATTR: {
+                            /* The object where the attribute is attached will be iterated */
+                            switch (loc_obj->u.attribute.parent_obj_type) {
+                                case H5I_FILE:
+                                case H5I_GROUP:
+                                    if (NULL == (iter_object = RV_group_open(loc_obj, loc_params, ".",
```
jhendersonHDF, line 809: Have you tested H5Ovisit on an attribute? It doesn't seem like these open calls would be opening the correct object and would instead be trying to open the attribute object as a group, dataset, datatype, etc.

src/rest_vol_object.c
```diff
@@ -578,10 +746,368 @@ RV_object_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_s
...
+
+                case H5I_DATASET:
+                case H5I_DATATYPE:
+                    /* Execute callback on object */
+
+                    if (RV_parse_response(response_buffer.buffer, NULL, &oinfo, RV_get_object_info_callback) <
+                        0)
+                        FUNC_GOTO_ERROR(H5E_LINK, H5E_CANTGET, FAIL, "can't iterate over links");
+
+                    if (!iter_object_name)
```
jhendersonHDF, line 1093: A few issues here:

- It looks like `iter_object_name` never gets set, although I might have missed it
- This looks like it will call the supplied callback twice on dataset and datatype objects if it manages to get past this check
- If the object is a dataset or datatype object, there's no need to make a curl request to try to get all the links from the object

src/rest_vol_object.c
mattjala: Good catch - the API tests don't try H5Ovisit on an attribute. I'll implement a test for that case, and have this make a request to server to get the actual object the attribute is attached to.

jhendersonHDF: It would probably be a good idea to keep track of the object that an attribute is attached to at create or open time if we don't already rather than adding another call to the server here

src/rest_vol_object.c
```diff
@@ -726,3 +1275,799 @@ RV_get_object_info_callback(char *HTTP_response, void *callback_data_in, void *c
...
+
+                /* Move the pointer from the "d" in "id" the start of the id itself in the response. */
+                link_id_ptr += 5;
+
+                char *id_end_ptr = strstr(link_id_ptr, "\"");
+
+                size_t id_len = (size_t)(id_end_ptr - link_id_ptr);
+
+                memcpy(table[i].object_URI, link_id_ptr, id_len);
+                table[i].object_URI[id_len] = '\0';
```
jhendersonHDF, line 1603: Is the above not something that can be done with YAJL before the NUL terminator is placed into the link subsection?

jhendersonHDF:
There are a couple remaining minor cleanup tasks, but I think in the interest of getting this merged the changes for `H5Ovisit` should be split into a separate PR, while the changes for `H5Oget_info_by_name/H5Oget_info_by_idx` should stay in this PR. The changes for `H5Ovisit` are fairly extensive and will take some time to review.

jhendersonHDF (approved):

PR #39, opened 2023-06-26
Implement H5Dwrite_multi / H5Dread_multi

mattjala:
These functions are tested under the new tests added in HDFGroup/vol-tests#57, which are run with the `H5VL_CAP_FLAG_DATASET_MORE` flag enabled. That flag isn't enabled for the REST VOL yet though, because other functions under it aren't implemented.

### DIFF
--- a/src/rest_vol.h
+++ b/src/rest_vol.h
@@ -84,6 +84,8 @@
 #define HTTP_SERVER_ERROR_MIN 500 /* Minimum and maximum values for the 500 class of */
 #define HTTP_SERVER_ERROR_MAX 599 /* HTTP server error responses */
 
+#define DEFAULT_POLL_TIMEOUT_MS 100
+
 /* Macros to check for various classes of HTTP response */
 #define HTTP_INFORMATIONAL(status_code)                                                                      \
     (status_code >= HTTP_INFORMATIONAL_MIN && status_code <= HTTP_INFORMATIONAL_MAX)
@@ -192,9 +194,36 @@
 #define CURL_PERFORM_NO_ERR(curl_ptr, ret_value)                                                             \
     CURL_PERFORM_INTERNAL(curl_ptr, FALSE, H5E_NONE_MAJOR, H5E_NONE_MINOR, ret_value)
 
-/* Helper macro to find the matching JSON '}' symbol for a given '{' symbol. This macro is
- * used to extract out all of the JSON within a JSON object so that processing can be done
- * on it.
+/* Counterpart of CURL_PERFORM that takes a response_buffer argument,
+ * instead of using the global response buffer.
+ * Currently not used. */
+#define CURL_PERFORM_NO_GLOBAL(curl_ptr, local_response_buffer, ERR_MAJOR, ERR_MINOR, ret_value)             \
+    CURL_PERFORM_INTERNAL_NO_GLOBAL(curl_ptr, response_buffer, TRUE, ERR_MAJOR, ERR_MINOR, ret_value)
+
+#define CURL_PERFORM_INTERNAL_NO_GLOBAL(curl_ptr, local_response_buffer, handle_HTTP_response, ERR_MAJOR,    \
+                                        ERR_MINOR, ret_value)                                                \
+    do {                                                                                                     \
+        CURLcode result = curl_easy_perform(curl_ptr);                                                       \
+                                                                                                             \
+        /* Reset the cURL response buffer write position pointer */                                          \
+        local_response_buffer.curr_buf_ptr = local_response_buffer.buffer;                                   \
+                                                                                                             \
+        if (CURLE_OK != result)                                                                              \
+            FUNC_GOTO_ERROR(ERR_MAJOR, ERR_MINOR, ret_value, "%s", curl_easy_strerror(result));              \
+                                                                                                             \
+        if (handle_HTTP_response) {                                                                          \
+            long response_code;                                                                              \
+                                                                                                             \
+            if (CURLE_OK != curl_easy_getinfo(curl_ptr, CURLINFO_RESPONSE_CODE, &response_code))             \
+                FUNC_GOTO_ERROR(ERR_MAJOR, ERR_MINOR, ret_value, "can't get HTTP response code");            \
+                                                                                                             \
+            HANDLE_RESPONSE(response_code, ERR_MAJOR, ERR_MINOR, ret_value);                                 \
+        } /* end if */                                                                                       \
+    } while (0)
+
+/* Helper macro to find the matching JSON '}' symbol for a given '{' symbol. This macro is                   \
+ * used to extract out all of the JSON within a JSON object so that processing can be done                   \
+ * on it.                                                                                                    \
  */
 #define FIND_JSON_SECTION_END(start_ptr, end_ptr, ERR_MAJOR, ret_value)                                      \
     do {                                                                                                     \
@@ -338,6 +367,11 @@ extern char curl_err_buf[];
  * cURL header list
  */
 extern struct curl_slist *curl_headers;
+
+/* Default initial size for the response buffer allocated which cURL writes
+ * its responses into
+ */
+#define CURL_RESPONSE_BUFFER_DEFAULT_SIZE 1024
 
 /*
  * Saved copy of the base URL for operating on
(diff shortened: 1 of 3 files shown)


jhendersonHDF:
It looks like this PR is trying to bring in the changes from the H5Ovisit PR

mattjala:
> It looks like this PR is trying to bring in the changes from the H5Ovisit PR

Fixed

src/rest_vol.h
```diff
@@ -192,9 +192,79 @@
...
+    do {                                                                                                     \
+        fprintf(stderr, "%zu transfers still running\n", still_running);                                     \
+        fprintf(stderr, "Performing...");                                                                    \
+        if (CURLM_OK != curl_multi_perform(curl_multi_ptr, &still_running))                                  \
+            FUNC_GOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, ret_value, "cURL multi perform error");             \
+                                                                                                             \
+        if (still_running)                                                                                   \
+            /* wait for activity, timeout or "nothing"  */                                                   \
+            fprintf(stderr, "Polling...");                                                                   \
+        if (CURLM_OK != curl_multi_poll(curl_multi_ptr, NULL, 0, 1000, NULL))                                \
```
jhendersonHDF, line 238: I'm guessing you want to skip the entirety of this and the code below if `!still_running`? Otherwise it looks like a `curl_multi_poll` call is still made even if the multi transfer is finished.

src/rest_vol.h
```diff
@@ -192,9 +192,79 @@
...
+/* Counterpart of CURL_PERFORM that takes a curl multi handle,
+ * and waits until all requests on it have finished before returning. */
+
+#define CURL_MULTI_PERFORM(curl_multi_ptr, ERROR_MAJOR, ERR_MINOR, ret_value)                                \
+    int      still_running = 0;                                                                              \
+    CURLMsg *curl_mult_msg = NULL;                                                                           \
+                                                                                                             \
+    do {                                                                                                     \
+        fprintf(stderr, "%zu transfers still running\n", still_running);                                     \
+        fprintf(stderr, "Performing...");                                                                    \
```
jhendersonHDF, line 231: It's fine to have these fprintf calls behind debug macros, but otherwise the REST VOL shouldn't print output at all except in cases where an error occurs and the error stack mechanism hasn't been disabled

src/rest_vol.h
```diff
@@ -620,6 +710,10 @@ herr_t RV_convert_dataspace_shape_to_JSON(hid_t space_id, char **shape_body, cha
 herr_t RV_base64_encode(const void *in, size_t in_size, char **out, size_t *out_size);
 herr_t RV_base64_decode(const char *in, size_t in_size, char **out, size_t *out_size);
 
+size_t H5_rest_curl_write_data_callback_no_global(char *buffer, size_t size, size_t nmemb, void *userp);
+
+size_t H5_rest_curl_write_data_callback_no_global(char *buffer, size_t size, size_t nmemb, void *userp);
```
jhendersonHDF, line 715: Duplicate prototype

src/rest_vol.h
```diff
@@ -379,6 +454,21 @@ extern const char *link_creation_time_keys[];
...
+extern const char *links_keys[];
+
+/* JSON keys to retrieve all of the information from a link when doing link iteration */
+extern const char *link_title_keys[];
+extern const char *link_creation_time_keys[];
+
+/* JSON keys to retrieve the collection that a hard link belongs to
+ * (the type of object it points to), "groups", "datasets" or "datatypes"
+ */
+extern const char *link_collection_keys2[];
```
jhendersonHDF, line 470: Looks like a holdout from the H5Ovisit PR, though it shouldn't affect anything

src/rest_vol.h
mattjala: This is an old macro that was unused at the time - I've since moved the finished version under the macro and removed these debug prints.

src/rest_vol.h
```diff
@@ -192,9 +192,198 @@
...
+    memset(failed_handles_to_retry, 0, sizeof(CURL *) * count);                                              \
+                                                                                                             \
+    do {                                                                                                     \
+        fail_count    = 0;                                                                                   \
+        succeed_count = 0;                                                                                   \
+                                                                                                             \
+        if (CURLM_OK != curl_multi_perform(curl_multi_handle, &num_still_running))                           \
+            FUNC_GOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "cURL multi perform error");                  \
+                                                                                                             \
+        if (CURLM_OK != curl_multi_poll(curl_multi_handle, NULL, 0, 0, &events_occurred))                    \
```
jhendersonHDF, line 253: Seems to me like we'd probably want to pass an actual timeout value here rather than 0. Docs on this suggest using at least 100ms, but also first checking if [curl_multi_timeout](https://curl.se/libcurl/c/curl_multi_timeout.html) returns a lower value than 100ms and if so, use that instead.

src/rest_vol.h
```diff
@@ -192,9 +192,198 @@
...
+                                                                                                             \
+                failed_handles_to_retry[i] = NULL;                                                           \
+            }                                                                                                \
+        }                                                                                                    \
+                                                                                                             \
+        /*                                                                                                   \
+        nanosleep(&delay, NULL);                                                                             \
+        */                                                                                                   \
+        num_prev_running = num_still_running;                                                                \
+    } while (num_still_running > 0);
```
jhendersonHDF, line 382: It's not a big deal at the moment, but I think this macro is at the point that it should just be a function. It's overly long for a macro and mixes in variables, memory management and more related to things that aren't passed in as a parameter and are assumed to be part of the function the macro is called in. For re-usability and readability, it should probably be refactored into a function at some point with necessary things being passed in. It's also usually better if the function allocating memory is the one freeing it, but that can be difficult in something like this where it's nice to free the memory as soon as we can.

src/rest_vol_dataset.c
```diff
@@ -416,207 +438,338 @@ RV_dataset_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name
  *              March, 2017
  */
 herr_t
-RV_dataset_read(size_t count, void *dset[], hid_t mem_type_id[], hid_t mem_space_id[], hid_t file_space_id[],
-                hid_t dxpl_id, void *buf[], void **req)
+RV_dataset_read(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_space_id[],
+                hid_t _file_space_id[], hid_t dxpl_id, void *buf[], void **req)
```
jhendersonHDF, line 442: Just an observation: with the large amount of things you're allocating in this function, the multi dataset path may be suffering from a lot of cache misses as you jump between the ~19-21 arrays allocated here. I believe it will be better for performance and much more readable to declare a structure type that has a field for everything you need related to an individual datasets and then allocate and free a `count`-sized array of those structures instead of having every array be separate in memory.

src/rest_vol_dataset.c
```diff
@@ -3724,8 +4049,120 @@ RV_convert_buffer_to_obj_refs(char *ref_buf, size_t ref_buf_len, rv_obj_ref_t **
...
+    response_read_info *resp_info = (response_read_info *)op_data;
+    *src_buf                      = resp_info->response_buf->buffer;
+    *src_buf_bytes_used           = *((size_t *)resp_info->read_size);
 
     return 0;
 } /* end dataset_read_scatter_op() */
+
+/* Return the index of the curl handle whose URL matches that of the given handle*/
+herr_t
+RV_get_index_of_matching_handle(CURL **curl_handles, size_t count, CURL *handle, size_t *handle_index)
```
jhendersonHDF, line 4061: What's the significance of needing to find a handle with a matching URL here? Since we're searching for a `CURL *` in an array of `CURL *`, can't you just compare the pointers for equality and then return the index? Or if not, create an array of structures where the structure holds a `CURL *` and any other information you need for quick matching later? I imagine this function could get pretty slow with even a smaller number of requests due to a potential worst-case of `count` string comparisons each time we need to find the index for a handle, which also appears to happen pretty often.

src/rest_vol.h
mattjala: Setting a timeout had an issue with always waiting that duration, regardless of activity. Switching to select() instead of poll() let it use the timeout correctly (see new commit).

jhendersonHDF: Yes, curl_multi_poll is blocking and so it's going to wait for handle activity or until the timeout has passed. With the new changes, it doesn't look like a call to `select()` was added, unless I missed it. But I think this is achieving the same thing either way, you're still going to wait on the handle for the given timeout that you give to the `select()` call. I think using `curl_multi_poll()` is just cleaner here. A decision needs to be made as to whether there is useful work to do if a 0 timeout is going to be used. Otherwise I think you want to use a small timeout like 100ms regardless to give curl time to process requests and not keeping making `curl_multi_poll` calls that return immediately after doing nothing.

mattjala: Looks like I accidentally deleted the select line before committing; it should be there now.

I tested the time `curl_multi_poll()` and the overall vol tests took with a few timeout values. With 0 ms the vol tests took about 12 seconds, while a 10 ms default timeout (lowered if _timeout returned a lower value) took it up to ~36 seconds. Checking each iteration showed that `curl_multi_poll()` was always taking at least 1 ms to wait. The documentation for poll says it should return before the timeout if activity is detected, so I suspect that something more needed to be setup for it to detect activity on the handles.

The new select() method waits for the timeout on some iterations, but returns near instantly on others, so the timeout seems to be returning early when activity is detected now.

jhendersonHDF:
It looks like all the changes from recent PRs made it into this PR from the merge. It's probably best to rebase it off of the master branch

jhendersonHDF (approved):

PR #41, opened 2023-07-12
Support H5Iget name

mattjala:
Associating a path/name with each open object goes against the spirit
of the API, but the alternative is traversing every link in the
domain, which is already slow in the library and would be
slower here.

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -2462,7 +2462,12 @@ RV_copy_object_loc_info_callback(char *HTTP_response, void *callback_data_in, vo
         new_domain->u.file.fcpl_id        = H5Pcopy(loc_info_out->domain->u.file.fcpl_id);
         new_domain->u.file.ref_count      = 1;
         new_domain->u.file.server_version = found_domain.u.file.server_version;
-        new_domain->handle_path           = "/";
+            
+        /* Allocate root "path" on heap for consistency with other RV_object_t types */
+        if ((new_domain->handle_path = RV_malloc(2)) == NULL)
+            FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTALLOC, NULL, "can't allocate space for filepath");
+
+        strncpy(new_domain->handle_path, "/", 2);
 
         /* Assume that original domain and external domain have the same server version.
          * This will always be true unless it becomes possible for external links to point to
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -3382,6 +3382,56 @@ done:
     return (ret_value);
 } /* end RV_set_object_type_header */
 
+/* Helper function to initialize an object's name based on its parent's name. 
+ * Allocates memory that must be closed by caller. */
+herr_t RV_set_object_handle_path(const char *obj_path, const char *parent_path, char **buf) {
+    herr_t ret_value = SUCCEED;
+    hbool_t include_parent_path = false;
+    size_t path_size = 0;
+    size_t path_len = 0;
+    char *handle_path = NULL;
+
+    /* Objects can be created/opened without reference to their path. Leave handle_path NULL in this case */
+    if (!obj_path) {
+        *buf = NULL;
+        FUNC_GOTO_DONE(SUCCEED);
+    }
+
+    /* Parent name is included if it is not the root and the object is opened by relative path */
+    include_parent_path = parent_path && strcmp(parent_path, "/") && (obj_path[0] != '/');
+
+    path_size = include_parent_path ? 
+        strlen(parent_path) + 1 + strlen(obj_path) + 1 
+        : 1 + strlen(obj_path) + 1;
+
+    if ((handle_path = RV_malloc(path_size)) == NULL)
+        FUNC_GOTO_ERROR(H5E_SYM, H5E_CANTALLOC, NULL, "can't allocate space for handle path");
+
+    if (include_parent_path) {
+        strncpy(handle_path, parent_path, strlen(parent_path));
+        path_len += strlen(parent_path);
+    }
+
+    /* Add leading slash if not in group name */
+    if (obj_path[0] != '/') {
+        handle_path[path_len] = '/';
+        path_len += 1;
+    }
+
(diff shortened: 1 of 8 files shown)


src/rest_vol.c
```diff
@@ -3380,6 +3387,57 @@ RV_set_object_type_header(H5I_type_t parent_obj_type, const char **parent_obj_ty
...
+    }
+
+    /* Parent name is included if it is not the root and the object is opened by relative path */
+    include_parent_path = parent_path && strcmp(parent_path, "/") && (obj_path[0] != '/');
+
+    path_size =
+        include_parent_path ? strlen(parent_path) + 1 + strlen(obj_path) + 1 : 1 + strlen(obj_path) + 1;
+
+    if ((handle_path = RV_malloc(path_size)) == NULL)
+        FUNC_GOTO_ERROR(H5E_SYM, H5E_CANTALLOC, NULL, "can't allocate space for handle path");
```
jhendersonHDF, line 3414: Should set ret_value to FAIL instead of NULL

src/rest_vol.c
```diff
@@ -3380,6 +3387,57 @@ RV_set_object_type_header(H5I_type_t parent_obj_type, const char **parent_obj_ty
...
+
+    if (include_parent_path) {
+        strncpy(handle_path, parent_path, strlen(parent_path));
```
jhendersonHDF, line 3417: Generally the third argument to strncpy should never depend on the second argument because that isn't really useful. You want the argument to be based off of some variable that stores that max size for the first argument, which in this case would be `path_size`. It's mostly harmless here since `handle_path` is allocated in the same function and we know how big the buffer is, but it's worth knowing about.

src/rest_vol.c
```diff
@@ -3380,6 +3387,57 @@ RV_set_object_type_header(H5I_type_t parent_obj_type, const char **parent_obj_ty
...
+
+    /* Add leading slash if not in group name */
+    if (obj_path[0] != '/') {
+        handle_path[path_len] = '/';
+        path_len += 1;
+    }
+
+    strncpy(handle_path + path_len, obj_path, strlen(obj_path) + 1);
+    path_len += (strlen(obj_path) + 1);
+
```
jhendersonHDF, line 3429: To generally be safe, it would be good to add a NUL terminator at `handle_path[path_size - 1]` before returning it

src/rest_vol_attr.c
```diff
@@ -114,6 +116,20 @@ RV_attr_create(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_
     new_attribute->domain = parent->domain;
     parent->domain->u.file.ref_count++;
 
+    new_attribute->handle_path = NULL;
+
+    if (RV_set_object_handle_path(attr_name, parent->handle_path, &new_attribute->handle_path) < 0)
```
jhendersonHDF, line 121: Since the documentation for `H5Iget_name` says that it returns the name of the object to which the attribute is attached, I don't think you want to add the attribute's name into `handle_path` here. In fact, you should just be able to return `parent->handle_path` when `H5Iget_name` is called on an attribute.

mattjala: This case is handled [here](https://github.com/HDFGroup/vol-rest/pull/41/files#diff-de2be4672626f5b68efea83e058e0d4b07beb9fd876ed4703032102ed3135e2eR308) in the logic for `H5Iget_name` itself. I think having the attribute store its name in its path is more consistent, and more flexible in case any other API functions work differently.

src/rest_vol_object.c
```diff
@@ -291,8 +291,37 @@ RV_object_get(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_get_ar
...
+                 * to */
+                if (loc_obj->obj_type == H5I_ATTR)
+                    if ((name = loc_obj->u.attribute.parent_name) == NULL)
+                        FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "attribute parent has NULL name");
+
+                copy_size = (strlen(name) < args->args.get_name.buf_size - 1)
+                                ? strlen(name)
+                                : args->args.get_name.buf_size - 1;
+                strncpy(args->args.get_name.buf, name, copy_size);
+                args->args.get_name.buf[copy_size + 1] = '\0';
```
jhendersonHDF, line 316: Should be `copy_size` instead of `copy_size + 1`, but the `memset` above took care of already putting the NUL terminator in the right place.

src/rest_vol_object.c
```diff
@@ -291,8 +291,37 @@ RV_object_get(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_get_ar
...
+            }
+
+            if (args->args.get_name.name_len) {
+                *args->args.get_name.name_len = strlen(name);
```
jhendersonHDF, line 320: Have you compared this value against the native library's behavior? The docs aren't very clear on whether the returned name length includes the NUL terminator, but at first glance it appears that the library is returning the length of the name + 1 for the NUL terminator.

src/rest_vol_object.c
```diff
@@ -790,20 +819,26 @@ RV_object_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_s
```
jhendersonHDF, line 819: To be safe, you should initialize `iter_object->handle_path = NULL` here so that the added free call in the `done:` section below doesn't accidentally try to free a random pointer if some code is later added between the allocation of `iter_object` and the allocation/setting of `iter_object->handle_path`.

src/rest_vol_object.c
mattjala: I just tested it, the library returns the length of the name without the NUL terminator. So this behavior matches.

jhendersonHDF: Hmm ok, looks like I was looking at the logic for `H5Rget_obj_name` which returns `strlen + 1`, however it also apparently adds in a leading "/" character, so that's likely what the +1 is accounting for.

jhendersonHDF (approved):

PR #42, opened 2023-07-18
Fix build_vol_cmake.sh

mattjala:
vol-tests' configuration requires HDF5 to be built, so the script
now configures and builds the REST VOL/HDF5 before
configuring/building vol-tests.

The instructions for manual build with cmake will
eventually need to be updated for anyone who wants
to build the vol-tests as well.

### DIFF
--- a/build_vol_cmake.sh
+++ b/build_vol_cmake.sh
@@ -41,7 +41,7 @@ CURL_DEBUG_OPT=
 MEM_TRACK_OPT=
 THREAD_SAFE_OPT=
 PREBUILT_HDF5_OPT=
-
+PREBUILT_HDF5_DIR=
 
 echo
 echo "*************************"
@@ -133,6 +133,7 @@ while getopts "$optspec" optchar; do
         ;;
     H)
         PREBUILT_HDF5_OPT="-DPREBUILT_HDF5_DIR=$OPTARG"
+        PREBUILT_HDF5_DIR="$OPTARG"
         echo "Set HDF5 install directory to: $OPTARG"
         echo
         ;;
@@ -173,7 +174,7 @@ if [ "$NPROCS" -eq "0" ]; then
     fi
 fi
 
-# Ensure that the HDF5 and VOL tests submodules get checked out
+# Ensure that the HDF5 submodule gets checked out
 if [ -z "$(ls -A ${SCRIPT_DIR}/${HDF5_DIR})" ]; then
     git submodule init
     git submodule update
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/vol-tests/CMakeLists.txt)
   )
   message(STATUS ${GIT_SUBMODULE_UPDATE_OUTPUT})
 endif()
-add_subdirectory(vol-tests)
+#add_subdirectory(vol-tests)
```
jhendersonHDF, line 30: This should probably be put behind a CMake option rather than just being disabled as an option

mattjala: The issue this PR was addressing occurs because the vol-tests' configuration depends on HDF5 already being built. Having the vol-tests' CML file invoked like this can only work if a prebuilt library is being used, otherwise I don't think there's a way to tell CMake to wait for HDF5 to get built - as far as I can tell, CMake's `add_custom_target`/dependency logic is for dependencies at _build_ time, not configure time.

That said, the CMake build script still builds the vol-tests from the submodule. This set of changes just means if someone is doing CMake manually, they'll need to invoke it once for the REST VOL/HDF5, and once for the vol-tests. Which isn't perfect, but we'll be using the API tests in the library itself soon anyway.

jhendersonHDF: If it isn't currently possible, the VOL tests CMakeLists code could be updated to handle this correctly, but I think the short of it is that we shouldn't be building HDF5 as part of the project in this manner anyway. Probably best at this point to switch over to using [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) or similar.

build_vol_cmake.sh
```diff
@@ -202,4 +203,28 @@ if [ "${CMAKE_GENERATOR}" = "Unix Makefiles" ]; then
...
+mkdir -p "${BUILD_DIR}/tests/vol-tests"
+cd "${BUILD_DIR}/tests/vol-tests"
+
+if [ -z "$PREBUILT_HDF5_DIR" ]; then
+    HDF5_INSTALL_DIR=$HDF5_DIR
+else
+    HDF5_INSTALL_DIR=$PREBUILT_HDF5_DIR
+fi
+
+CFLAGS="-D_POSIX_C_SOURCE=200809L" cmake -G "${CMAKE_GENERATOR}" -DHDF5_DIR=${HDF5_INSTALL_DIR} -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}" "${CONNECTOR_DEBUG_OPT}" "${CURL_DEBUG_OPT}" "${MEM_TRACK_OPT}" "${THREAD_SAFE_OPT}" "${SCRIPT_DIR}/test/vol-tests"
```
jhendersonHDF, line 222: We should probably revisit this when we switch to the API tests. I'm not sure that this flag should really be needed.

jhendersonHDF (approved):

Issue #44, opened 2023-07-28
Support H5Dset_extent


mattjala:

mattjala:
Implemented in #49

PR #49, opened 2023-08-04
Support H5Dset_extent

mattjala:
In order to keep track of whether two open datasets reference the same dataset, and should share information about their dataspace, I introduced `RV_type_info_array_g`, a global array keeping track of opened objects in the REST VOL, similar to the library's `H5I_type_info_array_g`.

 Addresses #44.

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -16,6 +16,7 @@
  */
 
 #include "rest_vol_dataset.h"
+#include <math.h>
 
 /* Set of callbacks for RV_parse_response() */
 static herr_t RV_parse_dataset_creation_properties_callback(char *HTTP_response, void *callback_data_in,
@@ -350,7 +351,7 @@ RV_dataset_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name
     rv_hash_table_iter_t iterator;
     rv_hash_table_iterate(type_info->table, &iterator);
     hid_t        matching_dspace = H5I_INVALID_HID;
-    RV_object_t *other_dataset   = H5I_INVALID_HID;
+    RV_object_t *other_dataset   = NULL;
 
 #ifdef RV_CONNECTOR_DEBUG
     printf("-> Received dataset open call with following parameters:\n");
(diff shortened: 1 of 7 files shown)


test/test_rest_vol.c
```diff
@@ -1473,8 +1473,6 @@ test_unused_file_API_calls(void)
 
         if (H5Fflush(file_id, H5F_SCOPE_GLOBAL) >= 0)
             TEST_ERROR
-        if (H5Fget_obj_count(file_id, H5F_OBJ_DATASET) >= 0)
```
mattjala, line 1476: This is supported now, so it's no longer expected to fail. Actual tests for it are in `h5vl_test`.

src/rest_vol_dataset.c
```diff
@@ -394,7 +405,23 @@ RV_dataset_open(void *obj, const H5VL_loc_params_t *loc_params, const char *name
 #endif
 
     /* Set up a Dataspace for the opened Dataset */
-    if ((dataset->u.dataset.space_id = RV_parse_dataspace(response_buffer.buffer)) < 0)
+
+    /* If this is another view of an already-opened dataset, make them share the same dataspace
+     * so that changes to it (e.g. resizes) are visible to both views */
+    while (rv_hash_table_iter_has_more(&iterator)) {
+        other_dataset = (RV_object_t *)rv_hash_table_iter_next(&iterator);
```
jhendersonHDF, line 412: This is fine for now functionality wise, but we need a better solution for this rather than iterating through the hash table on each dataset open. I'd suggest refactoring this whole PR later where rather than inserting each object pointer into a hash table for the ID type, we hash the object's URI and insert that hashed value into the hash table for the ID type. Then, on each dataset (of other object) open, you only need to hash the URI here and do a single lookup to see if an object with this URI is already open.

src/rest_vol_dataset.c
```diff
@@ -1383,11 +1425,192 @@ RV_dataset_specific(void *obj, H5VL_dataset_specific_args_t *args, hid_t dxpl_id
...
+
+            /* Disable use of Expect: 100 Continue HTTP response */
+            curl_headers = curl_slist_append(curl_headers, "Expect:");
+
+            if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_HTTPHEADER, curl_headers))
+                FUNC_GOTO_ERROR(H5E_SYM, H5E_CANTSET, FAIL, "can't set cURL HTTP headers: %s", curl_err_buf);
+            if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_URL, request_url))
+                FUNC_GOTO_ERROR(H5E_SYM, H5E_CANTSET, FAIL, "can't set cURL request URL: %s", curl_err_buf);
+
+            /* First, make a GET request for the dataset's dataspace to do some checks */
```
jhendersonHDF, line 1469: I'd assume we should already have this information locally as a result of the create/open request? Is there a case where we need updated information from the server?

jhendersonHDF (approved):

src/rest_vol_dataset.c
mattjala: This call (and the later once to get DCPL) were redundant - removed in #99

PR #50, opened 2023-08-23
rest_vol_dataset: (fix) Write hyperslab (memory)

jwsblokland:
- Improved the function RV_dataset_write() such it makes use of the function H5Dgather() to gather the data needed for writing a hyperslab (memory) to a file. This approach is similar to the one found in the function RV_dataset_read().
- Adapted the hyperslab and point selection test.

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -829,6 +829,13 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t mem_spac
             FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "memory datatype is invalid");
 
         write_body_len = (size_t)file_select_npoints * dtype_size;
+        if (NULL == (write_body = (char *)RV_malloc(write_body_len)))
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, FAIL,
+                            "can't allocate space for the 'write_body' values");
+        if (H5Dgather(mem_space_id[0], buf[0], mem_type_id[0], write_body_len, write_body, NULL, write_body) <
+            0)
+            FUNC_GOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "can't gather data to write buffer");
+        buf[0] = write_body;
     } /* end if */
     else {
         if (H5T_STD_REF_OBJ == mem_type_id[0]) {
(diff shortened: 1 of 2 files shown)


mattjala:
The CI failures are due to an edge case in random datatype testing trying to create a zero-length string, not an issue with this PR.

Otherwise, LGTM.

mattjala (dismissed):

src/rest_vol_dataset.c
```diff
@@ -829,6 +829,13 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t mem_spac
             FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "memory datatype is invalid");
 
         write_body_len = (size_t)file_select_npoints * dtype_size;
+        if (NULL == (write_body = (char *)RV_malloc(write_body_len)))
```
jhendersonHDF, line 832: It's probably worth only allocating this buffer and calling H5Dgather if the memory space wasn't set to H5S_ALL and has a non-contiguous selection in it. Otherwise, the H5Dgather call could potentially add overhead for no reason. Checking for H5S_ALL should be straightforward, but detecting a non-contiguous selection may take a little effort in the way of H5S calls.

I also noticed that the code below for point selections allocates the same `write_body` buffer. If I'm not mistaken, it looks like writing with a point selection could cause the memory allocated here to be leaked.

mattjala: Looks like point selection does cause a memory leak. From valgrind with the vol-tests:

```
==309052== 40 bytes in 1 blocks are definitely lost in loss record 2 of 7
==309052==    at 0x4848899: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==309052==    by 0x541B4B2: ???
==309052==    by 0x4B2EBA7: H5VL__dataset_write (H5VLcallback.c:2236)
==309052==    by 0x4B34565: H5VL_dataset_write_direct (H5VLcallback.c:2280)
==309052==    by 0x4906D3B: H5D__write_api_common (H5D.c:1331)
==309052==    by 0x4909DDF: H5Dwrite (H5D.c:1388)
==309052==    by 0x149AF1: test_write_dataset_small_point_selection (vol_dataset_test.c:5758)
==309052==    by 0x161B98: vol_dataset_test (vol_dataset_test.c:11257)
==309052==    by 0x10FC1E: vol_test_run (vol_test.c:118)
==309052==    by 0x10FC1E: main (vol_test.c:279)
```

jwsblokland: @jhendersonHDF and @mattjala, thanks for the review and reminding me that I can use valgrind.

The mentioned memory leak of `write_body` has been address and also the strange behavior of `test_write_dataset_small_point_selection` test. For the test, it turns out I forgot the modify the `H5Dwrite()` call by including a memory dataspace.

The only thing left to implement is checking for H5S_ALL and non-contiguous selection.

mattjala: I think you made a typo - did you mean to ask about the case where `H5S_ALL == mem_space_id` and `H5S_ALL != file_space_id`? If so, I believe that copying the selection from file space is the intended behavior of the library - see the second row of the mem_space_id/file_space_id table in the [documentation for H5Dwrite](https://docs.hdfgroup.org/hdf5/develop/group___h5_d.html#ga98f44998b67587662af8b0d8a0a75906).

jwsblokland: Ah yes, you are right. Thanks for pointing me to the documentation of H5Dwrite().

mattjala:
Moved to #57

PR #51, opened 2023-08-23
CMake: (feature+fix) HDF5 and CMake Config files

jwsblokland:
- Implemented support for making using of find_package(HDF5).
- Improved FindYAJL.cmake. Now one can select shared or static YAJL library version.
- Improved the CMake build system such proper hdf5_vol_rest-config.cmake, hdf5_vol_rest-config-version.cmake and hdf5_vol_rest-targets-*.cmake are generated. Also install the improved FindYAJL.cmake. This is all needed to more easily make use of VOL REST API in other CMake projects.
- Improved the README.md file.
- Improved the build_vol_cmake.sh file.


jwsblokland:
The modifications to the `rv-config.cmake.in` are done such it remains close to its original implementation. With this pull request there is an opportunity to revisit certain choices, like the target names.

jhendersonHDF (approved):

jhendersonHDF:
Thanks for the updates @jwsblokland!

> With this pull request there is an opportunity to revisit certain choices, like the target names.

Yes, the CMake code for building the REST VOL was hastily thrown together and so could likely use some updates. For VOL connectors developed by the HDF Group staff, we've mostly been moving to a naming scheme similar to "hdf5-vol-name" or "hdf5_vol_name", depending on the context, so it's probably best for us to stay consistent there. For an example, the DAOS VOL's CMake files are at https://github.com/HDFGroup/vol-daos/tree/master/CMake.

CMakeLists.txt
```diff
@@ -565,10 +587,15 @@ include (${HDF5_VOL_REST_RESOURCES_DIR}/HDFCompilerFlags.cmake)
 # Build HDF5, unless the user has specified to use
 # a pre-built HDF5 distribution
 #-----------------------------------------------------------------------------
-if (NOT PREBUILT_HDF5_DIR)
+if (NOT PREBUILT_HDF5_DIR AND NOT HDF5_FOUND)
   include_directories(${HDF5_VOL_REST_SRC_DIR}/${HDF5_DIR_NAME}/src)
   include_directories(${HDF5_VOL_REST_SRC_DIR}/${HDF5_DIR_NAME}/src/H5FDsubfiling)
   add_subdirectory(${HDF5_DIR} ${PROJECT_BINARY_DIR}/${HDF5_DIR_NAME})
+elseif (NOT PREBUILD_HDF5_DIR AND HDF5_FOUND)
```
jhendersonHDF, line 594: Looks like this should be `PREBUILT_HDF5_DIR` instead

jhendersonHDF:
It looks like these changes are causing the PREBUILT_HDF5_DIR builds to fail for some reason. Though, at this point I think the whole prebuilt hdf5 notion should just go away.

jwsblokland:
> It looks like these changes are causing the PREBUILT_HDF5_DIR builds to fail for some reason. Though, at this point I think the whole prebuilt hdf5 notion should just go away.

In my latest commit I got all 3 options (build from source, prebuilt and find_pacakage) of specifying the HDF5 library working. Furthermore, the compilation of the examples works again.

CMakeLists.txt
```diff
@@ -127,7 +138,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${HDF5_VOL_REST_RESOURCES_MOD_DIR}")
 find_package (YAJL REQUIRED)
 if (YAJL_FOUND)
   include_directories(${YAJL_INCLUDE_DIRS})
-  set (LINK_LIBS ${LINK_LIBS} ${YAJL_LIBRARIES})
+  set (LINK_LIBS ${LINK_LIBS} yajl)
```
mattjala, line 141: Should this be `${YAJL_LIBRARIES}` instead of `yajl` to match how curl is linked?

mattjala:
Trying this on my end and having issues linking to yajl when running the build script. Tried it with prebuilt/non-prebuilt HDF5, and yajl installed under `/usr/local/` on Ubuntu 22.04, with cmake 3.22.1

Excerpt from the CMake error log:

```
...
/usr/bin/ld: ../bin/libhdf5_vol_rest.so: undefined reference to `yajl_tree_get'
/usr/bin/ld: ../bin/libhdf5_vol_rest.so: undefined reference to `yajl_tree_parse'
/usr/bin/ld: ../bin/libhdf5_vol_rest.so: undefined reference to `yajl_tree_get'
[ 97%] Building C object test/CMakeFiles/test_rest_vol-shared.dir/test_rest_vol.c.o
collect2: error: ld returned 1 exit status
make[2]: *** [examples/CMakeFiles/rv_crtdat-shared.dir/build.make:100: bin/rv_crtdat-shared] Error 1
make[1]: *** [CMakeFiles/Makefile2:988: examples/CMakeFiles/rv_crtdat-shared.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
collect2: error: ld returned 1 exit status
collect2: error: ld returned 1 exit status
collect2: error: ld returned 1 exit status
make[2]: *** [examples/CMakeFiles/rv_write-shared.dir/build.make:100: bin/rv_write-shared] Error 1
make[2]: *** [examples/CMakeFiles/rv_crtgrpd-shared.dir/build.make:100: bin/rv_crtgrpd-shared] Error 1
make[1]: *** [CMakeFiles/Makefile2:1118: examples/CMakeFiles/rv_crtgrpd-shared.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:1170: examples/CMakeFiles/rv_write-shared.dir/all] Error 2
collect2: error: ld returned 1 exit status
...
```

mattjala:
Reinstalling yajl to `/usr/` instead of `/usr/local/` allowed me to configure and build.  I'm not sure why this would make the linker unable to find the library, since both of those are in the linker's default search paths.

mattjala:
Upon further testing, a fresh yajl installation in `/usr/` or `/usr/local/` works. The issue was probably some duplicate installations on my end conflicting. 

This works with the prebuilt library, HDF5 found by CMake, and using the submodule.  Static/shared yajl also both worked. LGTM.

mattjala (approved):

PR #53, opened 2023-08-24
Fix chance of bad type creation in CI

mattjala:


jhendersonHDF (approved):
We should really do away with test_rest_vol as soon as we can. Hopefully we can get the build process with the library working smoothly. Will test again with the merged cmake updates.

PR #56, opened 2023-08-30
Update CMake to build under library

mattjala:
Depends on the changes to CMakeVOL.cmake from HDFGroup/hdf5#3450

- REST VOL must now be built at library build time with CMake, as described in the README
- Removed build scripts, autotools files, and PREBUILT_HDF5_DIR options.
- Removed hdf5 and vol-tests submodules (recent additions and outstanding PRs to vol-tests will need to get merged to HDF5's API tests).
- CI workflow now uses ctest to run the `test_rest_vol` file with valgrind.
- README describes updated build process


src/CMakeLists.txt
```diff
@@ -109,15 +110,10 @@ endif ()
...
-  add_custom_target(
-    rest_vol_pubconf_depend
-    DEPENDS ${PREBUILT_HDF5_DIR}/include/H5pubconf.h
-  )
-elseif (NOT PREBUILT_HDF5_DIR AND HDF5_FOUND)
+if (HDF5_FOUND)
   add_custom_target(
     rest_vol_pubconf_depend
-    DEPENDS ${HDF5_INCLUDE_DIRS}/H5pubconf.h
+    DEPENDS ${HDF5_BINARY_DIR}/src/H5pubconf.h
```
jhendersonHDF, line 116: I still think we need a better solution than relying on variables that may or may not exist. The REST VOL won't always be built at the same time that HDF5 is and in the case where `HDF5_FOUND` is set from a `find_package` call where HDF5 was already installed on the system, `HDF5_BINARY_DIR` isn't going to exist.

mattjala: What is the range of build options we want to support? There's building under the library with FetchContent, building the REST VOL against a prebuilt HDF5 library, downloading the HDF5 source as a submodule, finding a pre-existing (but not necessarily prebuilt) HDF5 library via `find_package`, and then whatever equivalents these methods have in autotools.

jhendersonHDF: > building the REST VOL against a prebuilt HDF5 library
> finding a pre-existing (but not necessarily prebuilt) HDF5 library via find_package

These cases should be treated as exactly the same. At a minimum, we must support building the REST VOL against an HDF5 package that is already installed on the system (manually, via spack, via a package manager, etc.) as the default and preferred building method. Anything else should be seen as alternative conveniences for different purposes.

> There's building under the library with FetchContent

Since CMakeVOL.cmake sets the CMake variables that a `find_package(HDF5)` call would otherwise set, building the REST VOL under the library via FetchContent should work automatically with no special case handling as long as the REST VOL's CMake code does the standard thing of `find_package(HDF5)` -> link to `HDF5_LIBRARIES/HDF5_C_LIBRARIES` and include `HDF5_INCLUDE_DIRS`.
That should mean that on the particular line this comment is on, keeping the custom target as depending on `HDF5_INCLUDE_DIRS/H5pubconf.h` _should_ work fine, though if there's an extra directory we're missing or something like that, we should fix that.

> downloading the HDF5 source as a submodule

Downloading the HDF5 source as a submodule should go away with this PR, so we shouldn't have to worry about that case anymore; it was just a convenience for users at the time and also for development purposes. Now that the CMake FetchContent scheme is in place, though, that should likely be the preferred method for development.

> whatever equivalents these methods have in autotools

Since Autotools is less convenient in this regard, we should consider that case as being equivalent to "building the REST VOL against an HDF5 package that is already installed on the system"

configure.ac
jhendersonHDF, line 1: Same comment here about removing Autotools support

build_vol_autotools.sh
jhendersonHDF, line 1: At least parts of these script should go away, but the more I think about it the more that I realize these can probably stay in as convenience wrappers around the FetchContent building method. Probably a future improvement to look into.

The scripts can also still be useful for wrapping around the debug build options conveniently. For the time being, I think we should instead stick to just removing the prebuilt HDF5 parts of the scripts and leave the rest. The build system code will be responsible for locating HDF5 and we can still allow people to pass in a directory with the -H option if need be.

src/CMakeLists.txt
mattjala: It was my understanding the `find_package` is meant to be used with libraries that are already built/installed. I haven't been able to use it with the HDF5 source - in config mode, `find_package` tries to find the `hdf5-config.cmake` file that won't be created from `hdf5-config.cmake.in` until the HDF5 configuration step is done.  In module mode, the provided `FindHDF5.cmake` module behaves the same way when looking for the config file.

mattjala: >That should mean that on the particular line this comment is on, keeping the custom target as depending on HDF5_INCLUDE_DIRS/H5pubconf.h should work fine, though if there's an extra directory we're missing or something like that, we should fix that.

When HDF5 is built and not installed, `H5pubconf.h` is only found in `${HDF5_BINARY_DIR}/src/`, which isn't an include directory.  This becomes an issue when trying to build the REST VOL while the library is built and not installed, hence this change. We could potentially add `${HDF5_BINARY_DIR}/src/` to the list of include directories specified in `CMakeVOL.cmake`, but this seems like it will become confusing down the road.

jhendersonHDF: Yes, `find_package` is meant for looking for packages that are already installed on the system. See https://github.com/HDFGroup/hdf5/blob/develop/CMakeVOL.cmake#L150-L162 where the CMake code for building VOLs essentially rips out `find_package(HDF5)` calls from the downloaded local copy of a VOL connector. By doing this and setting the relevant CMake variables, it basically has the effect of presenting a not-yet-built HDF5 as an installed package so that the VOL can be configured, and then it builds the VOL after HDF5 gets built so that everything is available as it should be (minus the installed directories and their contents being available).

It's likely the case that we can remove the custom depends target now since IIRC it was mostly just for supporting the prebuilt HDF5 notion. Now, if you're building against a pre-installed HDF5 you don't need to depend on anything before building the VOL because HDF5 is already available. If you're using the FetchContent approach, CMake should be correctly handling the ordering so that the VOL isn't built before HDF5 is. I would try removing the custom target and see if everything builds correctly as is.

mattjala: You're right, removing the custom target didn't mess up the build. I'll set something up to handle building against not-yet-built HDF5 and update this PR.

CMakeLists.txt
```diff
@@ -587,27 +569,13 @@ include (${HDF5_VOL_REST_RESOURCES_DIR}/HDFCompilerFlags.cmake)
...
-       libhdf5.so
-       libhdf5_hl.so
-       libhdf5.a
-       libhdf5_hl.a
-  )
-endif ()
+include_directories(${HDF5_INCLUDE_DIRS})
+
+list (APPEND HDF5_LIBRARIES_TO_EXPORT
+  ${HDF5_C_HL_LIBRARIES}
```
jhendersonHDF, line 577: Looks like `HDF5_C_LIBRARIES` probably needs to be added back here, but also looks like the list being appended to should be `HDF5_VOL_REST_LIBRARIES_TO_EXPORT` instead of `HDF5_LIBRARIES_TO_EXPORT` since the latter would be something HDF5 would set

CMakeLists.txt
```diff
@@ -459,38 +454,25 @@ configure_file (${HDF5_VOL_REST_RESOURCES_DIR}/rv_cmake_config.h.in ${HDF5_VOL_R
...
-    ${HDF5_DIR}
-    ${HDF5_BINARY_DIR}
-    CACHE
-    INTERNAL
+)
+
+list(APPEND HDF5_VOL_REST_INCLUDE_DIRECTORIES
+  ${HDF5_INCLUDE_DIRS}
+  ${HDF5_BINARY_DIR}/src
+  ${HDF5_SOURCE_DIR}/${HDF5_HL_DIR_NAME}/src
```
jhendersonHDF, line 467: The references to `HDF5_BINARY_DIR` and `HDF5_SOURCE_DIR` should be removed here as they shouldn't be needed and depending on how the connector is set, won't be set to anything

build_vol_cmake.sh
```diff
@@ -70,6 +65,8 @@ usage()
     echo "              library should be built with position independent"
     echo "              code option enabled."
     echo
+    echo "      -x      Enable building of the REST VOL examples."
```
jhendersonHDF, line 68: We should have building of the examples on by default instead of needing a build option for them. If they don't currently compile, we should fix that.

src/CMakeLists.txt
```diff
@@ -90,8 +91,8 @@ endif ()
 if (BUILD_SHARED_LIBS)
   file (MAKE_DIRECTORY "${HDF5_VOL_REST_BINARY_DIR}/shared")
   add_library (${HDF5_VOL_REST_LIBSH_TARGET} SHARED ${HDF5_VOL_REST_SRCS} ${HDF5_VOL_REST_HDRS})
-  TARGET_C_PROPERTIES (${HDF5_VOL_REST_LIBSH_TARGET} SHARED " " " ")
-  target_link_libraries (${HDF5_VOL_REST_LIBSH_TARGET} PUBLIC ${HDF5_LIBRARIES_TO_EXPORT} ${LINK_SHARED_LIBS})
+  RV_TARGET_C_PROPERTIES (${HDF5_VOL_REST_LIBSH_TARGET} SHARED " " " ")
+  target_link_libraries (${HDF5_VOL_REST_LIBSH_TARGET} PUBLIC ${HDF5_LIBRARIES_TO_EXPORT} ${LINK_SHARED_LIBS} ${HDF5_LIBRARIES})
```
jhendersonHDF, line 95: `HDF5_LIBRARIES` should be added to the `target_link_libraries` call for `HDF5_VOL_REST_LIB_TARGET` as well.

jhendersonHDF, line 95: We should also probably replace all usages of `HDF5_LIBRARIES_TO_EXPORT` with `HDF5_VOL_REST_LIBRARIES_TO_EXPORT` in this project.

test/vol-tests
jhendersonHDF, line 1: The changes made to this submodule seems to be bringing in some commits that dont exist in the current head of the VOL tests repo. The commit it should be updated to should be `996dd87212b2547f1ce638d29b64c8ca436d859c`

.github/workflows/main.yml
jhendersonHDF, line 1: We should avoid any of the steps in this workflow file using valgrind to run tests since for the most part that will just make them take significantly longer. We can setup another workflow in the future that uses sanitizers to check for this stuff without affecting the runtime much.

jhendersonHDF (approved):

PR #57, opened 2023-08-30
Fix Write Hyperslab from Memory

mattjala:
This is #50 with an additional check to avoid allocating the buffer and performing H5Dgather if the memory selection is already contiguous.


jwsblokland:
@mattjala I just noticed your work on improving the memory-side hyperslab for writing. The interesting thing is I also just finished  a similar improvement. I will create a PR of it such you can take a look at it. Maybe we need to combine both PRs to get the best solution.

PR #58, opened 2023-08-31
rest_vol_dataset: (feature) Fixed memory-side hyperslab

jwsblokland:
- Improved memory-side hyperslab when writing data by taking into account if the selected data is contiguous and starting memory offset.
- Fixed various hyperslab and point tests.

This is improved version of PR #50. Related PR #57. Most likely a combination of PR #57 and #58 will result in the best solution.

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -35,11 +35,15 @@ static herr_t RV_convert_dataset_creation_properties_to_JSON(hid_t dcpl_id, char
 static herr_t RV_convert_dataspace_selection_to_string(hid_t space_id, char **selection_string,
                                                        size_t *selection_string_len, hbool_t req_param);
 
+/* Helper function for dataspace selection */
+static htri_t RV_dataspace_selection_is_contiguous(hid_t space_id);
+
 /* Conversion function to convert one or more rest_obj_ref_t objects into a binary buffer for data transfer */
-static herr_t RV_convert_obj_refs_to_buffer(const rv_obj_ref_t *ref_array, size_t ref_array_len,
-                                            char **buf_out, size_t *buf_out_len);
-static herr_t RV_convert_buffer_to_obj_refs(char *ref_buf, size_t ref_buf_len, rv_obj_ref_t **buf_out,
-                                            size_t *buf_out_len);
+static herr_t   RV_convert_obj_refs_to_buffer(const rv_obj_ref_t *ref_array, size_t ref_array_len,
+                                              char **buf_out, size_t *buf_out_len);
+static herr_t   RV_convert_buffer_to_obj_refs(char *ref_buf, size_t ref_buf_len, rv_obj_ref_t **buf_out,
+                                              size_t *buf_out_len);
+static hssize_t RV_convert_start_to_offset(hid_t space_id);
 
 /* Callbacks used for post-processing after a curl request succeeds */
 static herr_t rv_dataset_read_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf,
@@ -804,9 +808,11 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
     H5S_sel_type           sel_type = H5S_SEL_ALL;
     H5T_class_t            dtype_class;
     hbool_t                is_transfer_binary = FALSE;
+    htri_t                 contiguous         = FALSE;
     htri_t                 is_variable_str;
     hssize_t               mem_select_npoints  = 0;
     hssize_t               file_select_npoints = 0;
+    hssize_t               offset              = 0;
     size_t                 host_header_len     = 0;
     size_t                 write_body_len      = 0;
     size_t                 selection_body_len  = 0;
@@ -982,6 +988,25 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
                 FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "memory datatype is invalid");
 
             write_body_len = (size_t)file_select_npoints * dtype_size;
+            if ((contiguous = RV_dataspace_selection_is_contiguous(transfer_info[i].mem_space_id)) < 0)
+                FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_BADVALUE, FAIL,
+                                "Unable to determine if the dataspace selection is contiguous");
+            if (!contiguous) {
+                if (NULL == (transfer_info[i].u.write_info.write_body = (char *)RV_malloc(write_body_len)))
+                    FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, FAIL,
+                                    "can't allocate space for the 'write_body' values");
+                if (H5Dgather(transfer_info[i].mem_space_id, buf[i], transfer_info[i].mem_type_id,
+                              write_body_len, transfer_info[i].u.write_info.write_body, NULL,
+                              transfer_info[i].u.write_info.write_body) < 0)
+                    FUNC_GOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "can't gather data to write buffer");
+                buf[i] = transfer_info[i].u.write_info.write_body;
+            }
+            else {
+                if ((offset = RV_convert_start_to_offset(transfer_info[i].mem_space_id)) < 0)
+                    FUNC_GOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL,
+                                    "Unable to determine memory offset value");
+                buf[i] = buf[i] + offset * dtype_size;
+            }
         } /* end if */
         else {
             if (H5T_STD_REF_OBJ == transfer_info[i].mem_type_id) {
@@ -1060,6 +1085,8 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
                    transfer_info[i].u.write_info.base64_encoded_values);
 #endif
 
+            if (transfer_info[i].u.write_info.write_body)
+                RV_free(transfer_info[i].u.write_info.write_body);
             write_body_len = (strlen(fmt_string) - 4) + selection_body_len + value_body_len;
             if (NULL == (transfer_info[i].u.write_info.write_body = RV_malloc(write_body_len + 1)))
                 FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't allocate space for write buffer");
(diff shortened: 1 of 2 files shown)


src/rest_vol_dataset.c
```diff
@@ -3744,3 +3764,155 @@ dataset_read_scatter_op(const void **src_buf, size_t *src_buf_bytes_used, void *
...
+                /* Multiple blocks: count should be 1 except for the last dimension (fastest) */
+                for (i = 0; i < ndims - 1; i++) {
+                    if (count[i] > 1)
+                        FUNC_GOTO_DONE(false);
+                }
+            }
+
+            /* For contiguous, all faster running dimensions than the current dimension should be selected
+             * completely */
+            whole = (start[ndims - 1] == 0) && (count[ndims - 1] * block[ndims - 1] == dims[ndims - 1]);
```
mattjala, line 3846: This seems to require that the fastest running/last dimension be selected completely, which I don't think is necessary for the selection to be contiguous.

Should the comment say that all _slower_ running dimensions than the last dimension must be contiguous?

jwsblokland: I had the same thought yesterday until I did a small test which shows it was a wrong thought. Now I read contiguous selection as a selection which is contiguous in memory. If the fastest running/last dimension is not completely selected you need additional jumps to get the right location in memory. A way to visual this, at least for me, is by considering a 2D example and make various selections. For each selection translate it to which memory locations you need to get the selected data.

mattjala:
The current tests don't cover any cases where a selection is non-contiguous, or where the value returned from `RV_is_selection_contiguous` is nonzero. Once those cases are tested, this should be good to merge.

jwsblokland:
In my latest commit, I several improvements of the `RV_dataspace_selection_is_contiguous()` function. Now it captures also the errors correctly. Furthermore, I also implemented a memory-side hyperslab test in which the hyperslab is non-contiguous in memory.

src/rest_vol_dataset.c
```diff
@@ -829,6 +834,23 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t mem_spac
...
+                                "can't allocate space for the 'write_body' values");
+            if (H5Dgather(mem_space_id[0], buf[0], mem_type_id[0], write_body_len, write_body, NULL,
+                          write_body) < 0)
+                FUNC_GOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "can't gather data to write buffer");
+            buf[0] = write_body;
+        }
+        else {
+            if ((offset = RV_convert_start_to_offset(mem_space_id[0])) < 0)
+                FUNC_GOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL, "Unable to determine memory offset value");
+            buf[0] = buf[0] + offset * dtype_size;
```
mattjala, line 852: What's the motivation for this? I think library functions should already take this into account, since they'll have access to the memory dataspace, and this could potentially lead to issues if the library expects the empty space at the start of the buffer.

jwsblokland: In the case of binary transfer, the variable `buf[0]` is used to set `uinfo.buffer` and it assumes that is the starting location in memory of the data to be transfer. The amount of data that needs to be transfer is specified by `uinfo.buffer_size` which is set to `write_body_len`. In the case of contiguous dataspace selection this starting point does not have to be at the input memory location of `buf[0]` but has some offset to it. Therefore, one needs to compute this offset and add it to `buf[0]`. This is done all to avoid unnecessary overhead as requested by @jhendersonHDF in PR #50. A way to visual this case is to consider a 1D example and make a selection which is contiguous but does not start at the first memory position.

If desired I can modify one of the hyperslab tests such that the offset is non-zero.

mattjala: Ah, I see. It would be best to include a test with a nonzero offset, since this is new functionality.

jwsblokland:
I have implemented an additional test in which the memory hyperslab is continuous and the memory offset is non-zero.

Furthermore, I have re-implemented my changes based on the latest master branch. The main reason for this, is that due to the recent added support of multi-datasets it was easier to reimplement it than rebasing my old branch to the latest master.

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4164,157 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+        FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTGET, FAIL, "can't retrieve number of selected points");
+    if (npoints < 2)
+        FUNC_GOTO_DONE(TRUE);
+
+    if ((ndims = H5Sget_simple_extent_ndims(space_id)) < 0)
+        FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTGET, FAIL, "can't retrieve dataspace dimensionality");
+    if (!ndims)
+        FUNC_GOTO_DONE(TRUE);
+
+    if (H5S_SEL_HYPERSLABS == H5Sget_select_type(space_id)) {
```
mattjala, line 4204: This function will report that any point selection is contiguous.

mattjala:
Non-contiguous and nonzero offset cases look good. Should be good to merge once the point selection `RV_dataspace_selection_is_contiguous` issue is fixed.

src/rest_vol_dataset.c
jwsblokland: Yes, this is right. That was the original implementation.

Reading at the documentation of `H5Sselect_elements()` it is not complete clearly to me how it works. Looking at the given 1D example, should I read it like one selects 1st, 14th, 17th, 23rd and 8th element of 1D array which holds the data at these points. When calling `H5Dwrite()` this data array should have a length of at least 23 elements. If this is the correct way of looking at it then any point selection should be non-contiguous. In this case the correct data will be written to the right location. Of course it is sub-optimal for the case when the points are continuous in memory. I can implement this like this or is it desired to also check if the selected points are contiguous in memory to possible overhead caused by additional the memory copies.

mattjala: I believe you're right. When using a point selection, the memory buffer and memory dataspace should be contiguous (see Example 7 [here](https://davis.lbl.gov/Manuals/HDF5-1.8.7/UG/12_Dataspaces.html)) even if the file selection isn't, so this behavior makes sense.

mattjala (approved):

src/rest_vol_dataset.c
jwsblokland: I propose we do the following to resolve the point selection case:
1. Implement any point selection is non-contiguous such this PR can be merged.
2. Next, I create a separate PR in which I improve the point selection case by taking into account if it is continuous and if so compute the corresponding offset in memory. Similar as the hyperslab case.

mattjala: Good idea. I spoke with Jordan about this, and most point selections are not contiguous in memory, so it's safest to assume they are non-contiguous and perform the memory management in `H5Dgather`

One thing about your previous example - when calling `H5Dwrite()`, the data array would only need a length of 5 elements, to hold the 1st, 14th, 17th, 23rd, and 8th elements.

jwsblokland: Thanks for the information. It makes me realize that I have to take a closer at the point selection to get a better understanding how it actually works.

In my latest commit, I have implemented the assumption that any point selection is non-contiguous in memory.

test/test_rest_vol.c
```diff
@@ -8711,23 +8720,159 @@ test_write_dataset_data_verification(void)
...
 
     /* Write to first two rows of dataset */
+    mdims[0]  = dims[1] * 2;
+    start[0]  = 0;
+    stride[0] = 1;
+    count[0]  = dims[1] * 2;
+    block[0]  = 1;
+    if ((mspace_id = H5Screate_simple(1, mdims, NULL)) < 0)
+        TEST_ERROR
+    if (H5Sselect_hyperslab(mspace_id, H5S_SELECT_SET, start, stride, count, block) < 0)
```
jhendersonHDF, line 8730: Note that selecting a hyperslab in the memory space here isn't really necessary since it's just a contiguous block across all `dims[1] * 2` elements in the memory space that is created, but it's still ok to do so here.

test/test_rest_vol.c
```diff
@@ -8711,23 +8720,159 @@ test_write_dataset_data_verification(void)
...
+
+    data_size = dims[1] * 2 * DATASET_DATA_VERIFY_WRITE_TEST_DSET_DTYPESIZE;
+
+    if (NULL == (write_buf = malloc(data_size)))
+        TEST_ERROR
+
+    for (i = 0; i < dims[1]; i++)
+        ((int *)write_buf)[i] = 68;
+    for (i = dims[1]; i < data_size / DATASET_DATA_VERIFY_WRITE_TEST_DSET_DTYPESIZE; i++)
+        ((int *)write_buf)[i] = 67;
```
jhendersonHDF, line 8823: If I understand correctly, shouldn't the condition in the second for loop here also be `i < dims[1]` ? Otherwise it looks like the first loop sets half the elements in `write_buf` and then the second loop starts at an offset of `dims[1]` in `write_buf` and set the full number of elements, which would write `dims[1]` elements past the end of `write_buf`.

jhendersonHDF: Ah, my mistake, ignore. I see what you're doing here now.

src/rest_vol_dataset.c
```diff
@@ -1060,6 +1085,8 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
                    transfer_info[i].u.write_info.base64_encoded_values);
 #endif
 
+            if (transfer_info[i].u.write_info.write_body)
+                RV_free(transfer_info[i].u.write_info.write_body);
```
jhendersonHDF, line 1089: It would be good to set this pointer to NULL after freeing it so that if we fail to allocate it again below for whatever reason, the error cleanup code won't accidentally double-free the buffer.

src/rest_vol_dataset.c
```diff
@@ -982,6 +988,25 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
...
+            if ((contiguous = RV_dataspace_selection_is_contiguous(transfer_info[i].mem_space_id)) < 0)
+                FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_BADVALUE, FAIL,
+                                "Unable to determine if the dataspace selection is contiguous");
+            if (!contiguous) {
+                if (NULL == (transfer_info[i].u.write_info.write_body = (char *)RV_malloc(write_body_len)))
+                    FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, FAIL,
+                                    "can't allocate space for the 'write_body' values");
+                if (H5Dgather(transfer_info[i].mem_space_id, buf[i], transfer_info[i].mem_type_id,
+                              write_body_len, transfer_info[i].u.write_info.write_body, NULL,
+                              transfer_info[i].u.write_info.write_body) < 0)
```
jhendersonHDF, line 1000: Not an issue really, but you can probably pass NULL as the last parameter here since NULL was provided for H5Dgather's callback function.

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4164,165 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+        for (i = ndims - 2; i >= 0; i--) {
+            if ((dims[i] > 1) && (count[i] * block[i] > 1) && !whole)
+                FUNC_GOTO_DONE(FALSE);
+
+            whole = whole && (start[i] == 0) && (count[i] * block[i] == dims[i]);
+        }
+    } /* end if*/
+    else if (H5S_SEL_POINTS == H5Sget_select_type(space_id)) {
+        /* Assumption: any point selection is non-contiguous in memory */
+        FUNC_GOTO_DONE(FALSE);
```
jhendersonHDF, line 4257: It looks like this function mirrors what the library is currently doing (https://github.com/HDFGroup/hdf5/blob/develop/src/H5Spoint.c#L1817), so I'm content with this. I created https://github.com/HDFGroup/hdf5/discussions/3536 because I think this whole function is something that HDF5 can very easily expose and would save people from needing to spend effort on implementing this functionality when it already exists in the library and just isn't exposed publicly.

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4164,165 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+                            "can't allocate space for dimension 'dims' values");
+
+        if (H5Sget_simple_extent_dims(space_id, dims, NULL) < 0)
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTGET, -1, "can't get dataspace dimension size");
+
+        if (NULL == (start = (hsize_t *)RV_malloc((size_t)ndims * sizeof(*start))))
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, -1,
+                            "can't allocate space for hyperslab selection 'start' values");
+
+        if (H5Sget_regular_hyperslab(space_id, start, NULL, NULL, NULL) < 0)
```
jhendersonHDF, line 4308: Same comment here about checking for a regular hyperslab with `H5Sis_regular_hyperslab`. Alternatively, I think that `H5Sget_select_bounds` may still fit this use case and should work with both regular and non-regular hyperslabs so that you don't have to worry about how to handle non-regular hyperslabs.

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4164,165 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+
+        ret_value = start[0];
+        for (i = 1; i < ndims; i++) {
+            ret_value = ret_value * dims[i] + start[i];
+        }
+    } /* end if */
+    else if (H5S_SEL_POINTS == H5Sget_select_type(space_id)) {
+        FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_BADVALUE, -1,
+                        "for point selection, computing the offset is not supported");
+    } /* end else if */
```
jhendersonHDF, line 4319: May want to handle the other selection types (all, none, error, N) just for completeness

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4164,165 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+            if ((dims[i] > 1) && (count[i] * block[i] > 1) && !whole)
+                FUNC_GOTO_DONE(FALSE);
+
+            whole = whole && (start[i] == 0) && (count[i] * block[i] == dims[i]);
+        }
+    } /* end if*/
+    else if (H5S_SEL_POINTS == H5Sget_select_type(space_id)) {
+        /* Assumption: any point selection is non-contiguous in memory */
+        FUNC_GOTO_DONE(FALSE);
+    } /* end else if */
```
jhendersonHDF, line 4258: May want to handle the other selection types (all, none, error, N) just for completeness

jhendersonHDF:
@jwsblokland Thanks for all your work on this PR! I think it's very close with just a few minor comments on improvements that could be made.

test/test_rest_vol.c
jwsblokland: Ah, good to know this. In our code, we decided to always explicitly select a hyperslab to avoid the confusion which points have been selected.

test/test_rest_vol.c
jwsblokland: I am happy to ignore this comment.

src/rest_vol_dataset.c
jwsblokland: I have implemented this suggestion.

src/rest_vol_dataset.c
jwsblokland: Ah that is much better. I have implemented the change.

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4164,165 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, FAIL,
+                            "can't allocate space for hyperslab selection 'count' values");
+        if (NULL == (block = (hsize_t *)RV_malloc((size_t)ndims * sizeof(*block))))
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, FAIL,
+                            "can't allocate space for hyperslab selection 'block' values");
+
+        if (nblocks = H5Sget_select_hyper_nblocks(space_id) < 0)
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTGET, FAIL, "can't get number of hyperslab blocks");
+
+        if (H5Sget_regular_hyperslab(space_id, start, stride, count, block) < 0)
```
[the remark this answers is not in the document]
jwsblokland: For now, as suggested I implemented the check `H5Sis_regular_hyperslab()` and if it is a non-regular hyperslab the function will return it is non-contiguous in memory. Furthermore, it sounds like to properly support non-regular hyperslabs require much more work. I think it is better to create a separate PR for it.

src/rest_vol_dataset.c
jwsblokland: Ah that would be great improvement. In that case everyone benefit from it.

src/rest_vol_dataset.c
jwsblokland: This has been implemented.

src/rest_vol_dataset.c
jwsblokland: Good suggestion. These cases have been added.

src/rest_vol_dataset.c
jwsblokland: The other cases have been added.

jwsblokland:
@jhendersonHDF Similar as the improvement I made for ROS3, it was fun implementing this improvement. Again, I got a better understanding of dataspaces in HDF5. I addressed your comments. Regarding the support of non-regular hyperslabs, I think it is better to create a separate PR for it. Reading your comment about it, it sounds like it is not trivial at all.

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4165,201 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+
+        case H5S_SEL_POINTS:
+            /* Assumption: any point selection is non-contiguous in memory */
+            FUNC_GOTO_DONE(FALSE);
+            break;
+
+        case H5S_SEL_ALL:
+        case H5S_SEL_ERROR:
+        case H5S_SEL_N:
+        case H5S_SEL_NONE:
```
jhendersonHDF, line 4274: I'd prefer to get this PR merged, but if you could make the following changes in a future PR I'd appreciate it:

- H5S_SEL_ALL should return true for is_contiguous
- H5S_SEL_NONE should return false for is_contiguous
- H5S_SEL_ERROR and H5S_SEL_N should make this function return FAIL since it shouldn't get called with those

src/rest_vol_dataset.c
```diff
@@ -4137,4 +4165,201 @@ rv_dataset_write_cb(hid_t mem_type_id, hid_t mem_space_id, void *buf, struct res
...
+            FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_BADVALUE, -1,
+                            "for point selection, computing the offset is not supported");
+            break;
+
+        case H5S_SEL_ALL:
+        case H5S_SEL_ERROR:
+        case H5S_SEL_N:
+        case H5S_SEL_NONE:
+        default:
+            ret_value = 0;
```
jhendersonHDF, line 4353: Same comment here about H5S_SEL_ERROR and H5S_SEL_N making this function return a failure value

jhendersonHDF (approved):

Issue #59, opened 2023-08-31
H5Fis_accessible: reports error message


jwsblokland:
When I use the function `H5Fis_accessible()` to verify if my code can access certain file, I always the following error message like this

```
HDF5 REST VOL-DIAG: Error detected in HDF5 REST VOL (1.0.0) thread 1:
  #000: /glb/data/SeismicApps3/nljbl6/vol-rest/src/src/rest_vol_file.c line 400 in RV_file_open(): 410 - Resource has been deleted
    major: File accessibility
    minor: Unable to open file
```

This is a bit confusing. Originally, I thought I did something wrong but my code does exactly what it suppose to do for this case. Would it be possible to suppress this error message for `H5Dis_accessible()`? For now, I just ignore it, but I know for sure it will be confusing for other people when they start using my code in combination with HSDS.

mattjala:
Fixed in #60

PR #60, opened 2023-08-31
Quiet file open errors during H5Fis_accessible

mattjala:
Fix for #59

### DIFF
--- a/src/rest_vol_file.c
+++ b/src/rest_vol_file.c
@@ -617,13 +617,15 @@ RV_file_specific(void *obj, H5VL_file_specific_args_t *args, hid_t dxpl_id, void
             /* Initialize in case of failure */
             *ret_is_accessible = FALSE;
 
-            H5E_BEGIN_TRY {
+            H5E_BEGIN_TRY
+            {
                 file_id = H5Fopen(filename, H5F_ACC_RDWR, fapl_id);
-            } H5E_END_TRY;
+            }
+            H5E_END_TRY;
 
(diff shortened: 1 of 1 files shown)


src/rest_vol_file.c
```diff
@@ -616,11 +617,19 @@ RV_file_specific(void *obj, H5VL_file_specific_args_t *args, hid_t dxpl_id, void
             /* Initialize in case of failure */
             *ret_is_accessible = FALSE;
 
-            void *ret_file;
+            H5E_BEGIN_TRY
+            {
+                file_id = H5Fopen(filename, H5F_ACC_RDWR, fapl_id);
```
jhendersonHDF, line 622: Why the change to H5Fopen instead of RV_file_open? In general it's best to try to stick with minimal calling of public HDF5 APIs inside VOL connectors where possible, although that can be difficult at times.

mattjala: I thought this would be more consistent with other places in the VOL, but after checking I realize that this isn't the case. Is there a reason that it's better to avoid use of the external API? I would expect it to be better practice to use it, since internal VOL functions often depend on things the library has set up.

jhendersonHDF: Calling public API functions can currently mess with HDF5's API contexts, error stacks, and other things. But yes it's not always possible to avoid, especially for things like needing to call H5T routines.

jhendersonHDF (approved):

PR #61, opened 2023-08-31
Implement H5Fget_obj_ids

mattjala:


src/rest_vol_file.c
```diff
@@ -544,9 +551,44 @@ RV_file_get(void *obj, H5VL_file_get_args_t *args, hid_t dxpl_id, void **req)
 
         /* H5Fget_obj_ids */
         case H5VL_FILE_GET_OBJ_IDS:
-            FUNC_GOTO_ERROR(H5E_FILE, H5E_UNSUPPORTED, FAIL, "H5Fget_obj_ids is unsupported");
-            break;
 
+            if (args->args.get_obj_ids.max_objs <= 0)
```
jhendersonHDF, line 555: 0 is valid for `max_objs` and should just be a no-op in that case

src/rest_vol_file.c
```diff
@@ -544,9 +551,44 @@ RV_file_get(void *obj, H5VL_file_get_args_t *args, hid_t dxpl_id, void **req)
...
+                if (H5Iiterate(H5I_DATATYPE, RV_iterate_copy_hid_cb, &id_list) < 0)
+                    FUNC_GOTO_ERROR(H5E_FILE, H5E_OBJECTITERERROR, FAIL, "can't iterate over datatype ids");
+
+            if (args->args.get_obj_ids.types & H5F_OBJ_DATASET)
+                if (H5Iiterate(H5I_DATASET, RV_iterate_copy_hid_cb, &id_list) < 0)
+                    FUNC_GOTO_ERROR(H5E_FILE, H5E_OBJECTITERERROR, FAIL, "can't iterate over dataset ids");
+
+            if (args->args.get_obj_ids.types & H5F_OBJ_ATTR)
+                if (H5Iiterate(H5I_ATTR, RV_iterate_copy_hid_cb, &id_list) < 0)
+                    FUNC_GOTO_ERROR(H5E_FILE, H5E_OBJECTITERERROR, FAIL, "can't iterate over attribute ids");
```
jhendersonHDF, line 589: After all the H5I calls have been made, the value of `id_list.obj_count` needs to be assigned back to `args->args.get_obj_ids.count` so that the value is properly returned.

src/rest_vol_file.c
```diff
@@ -755,3 +797,28 @@ RV_file_close(void *file, hid_t dxpl_id, void **req)
...
+ * Purpose:     Callback for H5Iiterate() that copies the hid_t
+ *              of each library object into a list.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  Matthew Larson
+ *              August, 2023
+ */
+herr_t
+RV_iterate_copy_hid_cb(hid_t obj_id, void *udata)
```
jhendersonHDF, line 813: This callback needs to take `max_objs` into account so that it doesn't overrun the buffer given.

src/rest_vol_file.c
```diff
@@ -544,9 +552,50 @@ RV_file_get(void *obj, H5VL_file_get_args_t *args, hid_t dxpl_id, void **req)
...
+
+            if (requested_types & H5F_OBJ_DATASET)
+                if (H5Iiterate(H5I_DATASET, RV_iterate_copy_hid_cb, &id_list) < 0)
+                    FUNC_GOTO_ERROR(H5E_FILE, H5E_OBJECTITERERROR, FAIL, "can't iterate over dataset ids");
+
+            if (requested_types & H5F_OBJ_ATTR)
+                if (H5Iiterate(H5I_ATTR, RV_iterate_copy_hid_cb, &id_list) < 0)
+                    FUNC_GOTO_ERROR(H5E_FILE, H5E_OBJECTITERERROR, FAIL, "can't iterate over attribute ids");
+
+            args->args.get_obj_ids.count = id_list.obj_count;
```
jhendersonHDF, line 596: Needs to be `*args->args.get_obj_ids.count`

src/rest_vol_file.c
```diff
@@ -755,3 +804,31 @@ RV_file_close(void *file, hid_t dxpl_id, void **req)
...
+ * Programmer:  Matthew Larson
+ *              August, 2023
+ */
+herr_t
+RV_iterate_copy_hid_cb(hid_t obj_id, void *udata)
+{
+    herr_t          ret_value = SUCCEED;
+    object_id_list *id_list   = (object_id_list *)udata;
+
+    if (id_list->obj_count >= id_list->max_obj_count)
```
jhendersonHDF, line 825: This function shouldn't fail when `obj_count >= max_obj_count`. Rather, the callback should check that `obj_count < max_obj_count` and if so, do

    id_list->obj_id_list[id_list->obj_count] = obj_id;
    id_list->obj_count++;

Otherwise, if `obj_count >= max_obj_count` the callback should just return `H5_ITER_STOP` for short-circuit success. To continue, the function should return `H5_ITER_CONT` instead of `SUCCEED`, but that's just a minor detail since they have the same value anyway. You can look at https://github.com/HDFGroup/vol-daos/blob/master/src/daos_vol_file.c#L3928-L3962 for how the DAOS VOL implements it, though you can probably skip the part about checking the connector name.

jhendersonHDF (approved):

PR #63, opened 2023-09-06
Implement H5Fflush

mattjala:
Ignores the H5F_SCOPE_GLOBAL/H5F_SCOPE_LOCAL parameter, since file mounting isn't implemented in HSDS.

### DIFF
--- a/test/test_rest_vol.c
+++ b/test/test_rest_vol.c
@@ -1471,8 +1471,6 @@ test_unused_file_API_calls(void)
         hid_t               obj_id;
         void               *file_handle;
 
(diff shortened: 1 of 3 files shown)


test/test_rest_vol.c
```diff
@@ -1471,8 +1471,6 @@ test_unused_file_API_calls(void)
         hid_t               obj_id;
         void               *file_handle;
 
-        if (H5Fflush(file_id, H5F_SCOPE_GLOBAL) >= 0)
```
mattjala, line 1474: I tested it with the flush tests in `vol-tests` to make sure it functions correctly.

jhendersonHDF (approved):

PR #64, opened 2023-09-06
Check access plist on obj create/open

### LINKED ISSUE #46: CTest: h5vL_test - Attribute test failed
When I build REST VOL API using CMake and run ctest, the test h5vl_test fails. Looking at the verbose output of this test, the following attribute test has failed:
```
Testing H5Aopen with an invalid AAPL                                *FAILED*
```

Currently, I am not sure why this attribute tests fails.

mattjala:
Should fix #46, although  I wasn't able to replicate the issue on my end.


jhendersonHDF (approved):

jhendersonHDF:
Note that this was probably fixed in HDF5 itself in https://github.com/HDFGroup/hdf5/pull/2712. Might be that the version of HDF5 used when #46 was created didn't have that fix yet.

PR #66, opened 2023-09-11
Implement H5Adelete_by_idx

mattjala:


src/rest_vol_attr.c
```diff
@@ -1884,7 +1974,8 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
 
                 /* H5Adelete_by_idx */
                 case H5VL_OBJECT_BY_IDX: {
-                    FUNC_GOTO_ERROR(H5E_ATTR, H5E_UNSUPPORTED, FAIL, "H5Adelete_by_idx is unsupported");
+                    FUNC_GOTO_ERROR(H5E_ATTR, H5E_UNSUPPORTED, FAIL,
+                                    "invalid location parameters - this message should not appear!");
```
mattjala, line 1978: `H5VL_ATTR_DELETE_BY_IDX` has to be handled in a different top-level branch, and uses `H5VL_OBJECT_BY_NAME` as the type for its location parameters (see [here](https://github.com/HDFGroup/hdf5/blob/a0edeee49f77a1cb59c7969a451164dd595e0d01/src/H5A.c#L2183)), so this branch can't be used, and it shouldn't even be possible for a user to elicit this error.

src/rest_vol_attr.c
```diff
@@ -1614,7 +1614,7 @@ RV_attr_get(void *obj, H5VL_attr_get_args_t *args, hid_t dxpl_id, void **req)
                     char       parent_obj_URI[URI_MAX_LENGTH];
 
                     /* Retrieve the type and URI of the object that the attribute is attached to */
-                    search_ret = RV_find_object_by_path(loc_obj, loc_params->loc_data.loc_by_name.name,
```
jhendersonHDF, line 1617: The VOL passes the name of the object to start from in `loc_params->loc_data.loc_by_name`, so the old way was correct.

jhendersonHDF (approved):

PR #70, opened 2023-09-14
Github actions: (fix) valgrind

jwsblokland:
- Update APT sources before installing valgrind.


jhendersonHDF (approved):

jhendersonHDF:
I'd prefer if we didn't run tests in the actions using valgrind in the first place, but hopefully this will fix the current testing failures at least

PR #71, opened 2023-09-18
Fix point selection I/O for simple numeric types

mattjala:
- JSON values from point reads are now converted to binary to be properly scattered by H5Dscatter().
- Changed a couple symbols in base64 encoding alphabet to make it compliant with RFC 3548, since HSDS was confused by the presence of dashes in encodings.
- Fixed some warnings from implicit type conversion.

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -21,6 +21,9 @@
 static herr_t RV_parse_dataset_creation_properties_callback(char *HTTP_response, void *callback_data_in,
                                                             void *callback_data_out);
 
+static herr_t RV_json_values_to_binary_callback(char *HTTP_response, void *callback_data_in,
+                                                void *callback_data_out);
+
 /* Helper functions for creating a Dataset */
 static herr_t RV_setup_dataset_create_request_body(void *parent_obj, const char *name, hid_t type_id,
                                                    hid_t space_id, hid_t lcpl_id, hid_t dcpl,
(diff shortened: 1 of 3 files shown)


src/rest_vol_dataset.c
```diff
@@ -4363,3 +4393,111 @@ RV_convert_start_to_offset(hid_t space_id)
...
+#ifdef RV_CONNECTOR_DEBUG
+    printf("-> Converting response JSON values to binary buffer\n\n");
+#endif
+
+    if (!HTTP_response)
+        FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "HTTP response buffer was NULL");
+    if (!out_buf)
+        FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "output buffer was NULL");
+
+    if ((dtype_size = H5Tget_size(dtype_id)) <= 0)
```
jhendersonHDF, line 4433: Comparing this < 0 will probably generate a warning on some compilers since it's unsigned

src/rest_vol_dataset.c
```diff
@@ -4363,3 +4393,111 @@ RV_convert_start_to_offset(hid_t space_id)
...
+ *
+ *              Expects data in to be a pointer to an hid_t for the
+ *              datatype in the response, and data out to be the
+ *              address of a pointer that will point to the
+ *              newly allocated buffer.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ */
+static herr_t
+RV_json_values_to_binary_callback(char *HTTP_response, void *callback_data_in, void *callback_data_out)
```
jhendersonHDF, line 4412: This function's fine for a first cut, but we'll definitely need much more sophisticated parsing eventually because it's likely that we're going to get many other things than just `H5T_NATIVE_INT/FLOAT/DOUBLE` for the mem type

jhendersonHDF (approved):

PR #72, opened 2023-09-21
Implement type conversion for simple numeric types

mattjala:
Several of the helpers here are copied from [the DAOS VOL's datatype code](https://github.com/HDFGroup/vol-daos/blob/master/src/daos_vol_type.c).

Tests for this are implemented in HDFGroup/vol-tests#62.

Fix for #6

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -24,6 +24,9 @@ static herr_t RV_parse_dataset_creation_properties_callback(char *HTTP_response,
 static herr_t RV_json_values_to_binary_callback(char *HTTP_response, void *callback_data_in,
                                                 void *callback_data_out);
 
+/* Internal helper for RV_json_values_to_binary_callback */
+herr_t RV_json_values_to_binary_recursive(yajl_val value_entry, hid_t dtype_id, void *value_buffer);
+
 /* Helper functions for creating a Dataset */
 static herr_t RV_setup_dataset_create_request_body(void *parent_obj, const char *name, hid_t type_id,
                                                    hid_t space_id, hid_t lcpl_id, hid_t dcpl,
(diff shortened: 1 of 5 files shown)


src/rest_vol_dataset.c
```diff
@@ -982,14 +997,48 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
...
+            if ((mem_type_size = H5Tget_size(transfer_info[i].mem_type_id)) == 0)
+                FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "unable to get size of memory datatype");
+
+            /* Initialize type conversion */
+            RV_tconv_init(transfer_info[i].mem_type_id, &mem_type_size, transfer_info[i].file_type_id,
+                          &file_type_size, (size_t)file_select_npoints, TRUE, FALSE,
+                          &transfer_info[i].tconv_buf, &transfer_info[i].bkg_buf, NULL, &fill_bkg);
+
+            /* Perform type conversion on response values */
+            memset(transfer_info[i].tconv_buf, 0, file_type_size * (size_t)mem_select_npoints);
```
jhendersonHDF, line 1021: Since TRUE is passed above for `clear_tconv_buf`, there's no need to memset it here since it will be allocated with calloc

src/rest_vol_dataset.c
```diff
@@ -4526,3 +4606,90 @@ RV_json_values_to_binary_callback(char *HTTP_response, void *callback_data_in, v
...
+
+            if ((member_size = H5Tget_size(member_dtype_id)) == 0)
+                FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_CANTGET, FAIL, "can't get size of member datatype");
+
+            if ((member_val = YAJL_GET_OBJECT(value_entry)->values[i]) == NULL)
+                FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_PARSEERROR, FAIL,
+                                "failed to parse member of compound type");
+
+            if (RV_json_values_to_binary_recursive(member_val, member_dtype_id,
+                                                   (void *)((char *)value_buffer + offset)) < 0)
```
jhendersonHDF, line 4680: This will probably work for packed compound types, but if the compound datatype has padding in it in the same way that the natural C struct likely does, this will be problematic. Best to use H5Tget_member_offset to get the member offsets instead of assuming they're linearly next to each other

jhendersonHDF (approved):
Merging this for the functionality it brings, but there are a few minor improvements that can be made in the future.

PR #75, opened 2023-09-27
Fix bad reference counting in attr iteration

mattjala:
This could sometimes cause segfaults when trying to close groups. h5py's attribute tests specifically ran into this issue, since h5py opens a group in order to iterate over its attributes.

### DIFF
--- a/src/rest_vol_attr.c
+++ b/src/rest_vol_attr.c
@@ -2313,12 +2313,25 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
                         case H5I_FILE:
                             /* Copy fapl, fcpl, and filepath name to new object */
 
-                            if (H5I_INVALID_HID ==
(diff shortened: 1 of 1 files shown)


src/rest_vol_attr.c
```diff
@@ -2351,6 +2358,10 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
                                 FUNC_GOTO_ERROR(H5E_ATTR, H5E_CANTINC, FAIL,
                                                 "can't increment field's ref. count for copy of attribute's "
                                                 "parent datatype");
+                            if (H5Iinc_ref(loc_obj->u.datatype.tapl_id) < 0)
```
jhendersonHDF, line 2361: We should probably add if blocks here as well to check `loc_obj->u.datatype.tcpl_id != H5P_DATATYPE_CREATE_DEFAULT` and `loc_obj->u.datatype.tapl_id != H5P_DATATYPE_ACCESS_DEFAULT` before incrementing the ref. count for them

jhendersonHDF (approved):

PR #81, opened 2023-10-05
Set connection info even with non-default FAPL

mattjala:
h5py uses a dynamically loaded connector while providing a non-default FAPL, and still needs the connection info to be set.

I tested this with a manually linked program, and didn't find any issues.

### DIFF
--- a/src/rest_vol_file.c
+++ b/src/rest_vol_file.c
@@ -64,15 +64,8 @@ RV_file_create(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id, h
     printf("     - Default FAPL? %s\n\n", (H5P_FILE_ACCESS_DEFAULT == fapl_id) ? "yes" : "no");
 #endif
 
-    /*
-     * If the connector has been dynamically loaded, the FAPL used for
-     * creating the file will be a default FAPL, so we need to ensure
-     * that the connection information gets set.
-     */
-    if (fapl_id == H5P_FILE_ACCESS_DEFAULT)
-        if (H5_rest_set_connection_information() < 0)
-            FUNC_GOTO_ERROR(H5E_FILE, H5E_CANTINIT, NULL,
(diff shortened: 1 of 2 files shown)


src/rest_vol_file.c
```diff
@@ -64,15 +64,8 @@ RV_file_create(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id, h
     printf("     - Default FAPL? %s\n\n", (H5P_FILE_ACCESS_DEFAULT == fapl_id) ? "yes" : "no");
 #endif
 
-    /*
```
jhendersonHDF, line 67: Where were you running into an issue with this? If the connector is loaded dynamically, the file access property list should be `H5P_FILE_ACCESS_DEFAULT`, meaning that we need to ensure the connection information gets set here on file create and open. Otherwise, if the connector is not loaded dynamically, it had to have been set on a non-default file access property list via `H5Pset_fapl_rest_vol`, which calls `H5_rest_set_connection_information` unconditionally. This looks like in the non-default FAPL case, the connection information will get set twice.

mattjala: When dynamically loading the VOL, h5py uses a non-default FAPL and still ends up executing VOL callbacks. This check prevented the VOL from being initialized. The first specific error this led to was file creation failing upon trying to copy a property list, due to not finding it in the library's global type array. 

As is, the info will get set multiple times - we could have a global similar to `H5_rest_initialized_g` to track whether it's already been set and avoid repeating it.

jhendersonHDF: Ah ok. Yes it would be good if we had some mechanism to avoid setting that information twice. I don't think it will cause any issues currently, but it's good for future-proofing. It may be a bit tricky though because we need to keep track of the status on a per-file basis rather than globally since we may want to be able to open different files using different connection information.

jhendersonHDF (approved):

Issue #80, opened 2023-10-05
Cannot open datasets with fill values set in DCPL


mattjala:
An unsupported error is still thrown when a fill value is detected in `RV_parse_dataset_creation_properties_callback()`

mattjala:
Fixed in #83.

PR #83, opened 2023-10-09
Support opening a dataset with a fill value

mattjala:
Fix for #80 

Tests for this are in HDFGroup/vol-tests#63

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -1510,6 +1510,7 @@ RV_parse_dataset_creation_properties_callback(char *HTTP_response, void *callbac
 {
     yajl_val parse_tree         = NULL, creation_properties_obj, key_obj;
     hid_t   *DCPL               = (hid_t *)callback_data_out;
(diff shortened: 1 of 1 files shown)


src/rest_vol_dataset.c
```diff
@@ -1709,10 +1711,29 @@ RV_parse_dataset_creation_properties_callback(char *HTTP_response, void *callbac
      *                                                                            *
      ******************************************************************************/
     if ((key_obj = yajl_tree_get(creation_properties_obj, fill_value_keys, yajl_t_any))) {
-        /* TODO: Until fill value support is implemented, just push an error to the stack but continue ahead
-         */
-        FUNC_DONE_ERROR(H5E_DATASET, H5E_UNSUPPORTED, SUCCEED,
-                        "warning: dataset fill values are unsupported");
+        hid_t  fill_type               = H5I_INVALID_HID;
```
jhendersonHDF, line 1714: `fill_type` should be moved to the top of the function scope and then closed at the end of this function with `H5Tclose` if `fill_type != H5I_INVALID_HID`.

PR #85, opened 2023-10-10
Silence compiler warnings

mattjala:
Mostly dealing with issues around the `const` qualifier. Some changes of note:
- The data in field for `RV_parse_callback` now has the `const` qualifier, so most callbacks now have a `const` qualifier on their data in field as well.
- Removed the use of a callback in `RV_multi_perform`, now just checks the transfer type and uses the read callback directly. Removed the empty write callback. 
-  Because the `get_link_val_callback` needed to modify the size of the in buffer,the size of the in buffer was paired with the buffer itself in a new struct, `get_link_val_out`.
- Renamed `RV_parse_type -> RV_parse_object_class` to distinguish it from `RV_parse_datataype`

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -171,7 +171,7 @@ static size_t H5_rest_curl_write_data_callback(char *buffer, size_t size, size_t
 static char *H5_rest_url_encode_path(const char *path);
 
 /* Helper function to parse an object's type from server response */
-herr_t RV_parse_type(char *HTTP_response, void *callback_data_in, void *callback_data_out);
+herr_t RV_parse_object_class(char *HTTP_response, const void *callback_data_in, void *callback_data_out);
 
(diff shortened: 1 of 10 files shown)


src/rest_vol.c
```diff
@@ -1539,7 +1539,8 @@ H5_rest_url_encode_path(const char *_path)
     if (!_path)
         FUNC_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, NULL, "path was NULL");
 
-    path = (char *)_path;
+    /* Silence compiler const warnings */
+    memcpy(&path, &_path, sizeof(char *));
```
jhendersonHDF, line 1543: This works (if a bit awkward), although we could probably get away with just making `path` and `cur_pos` const while using a different var. than `cur_pos` when indexing into `tmp_buffer`.

jhendersonHDF (approved):

PR #86, opened 2023-10-12
Make server connection info file-local

mattjala:
`H5Fdelete` doesn't receive any information about the file to delete besides its filename, so this may be impossible to completely refactor. 

The global info (username/password in the global curl handle, global `base_URL` variable) haven't been removed yet, but each request to server now uses a file-local username and password.

### DIFF
--- a/src/rest_vol_link.c
+++ b/src/rest_vol_link.c
@@ -336,7 +336,12 @@ RV_link_create(H5VL_link_create_args_t *args, void *obj, const H5VL_loc_params_t
     uinfo.buffer      = create_request_body;
     uinfo.buffer_size = (size_t)create_request_body_len;
     uinfo.bytes_sent  = 0;
-
+    if (CURLE_OK !=
+        curl_easy_setopt(curl, CURLOPT_USERNAME, new_link_loc_obj->domain->u.file.server_info.username))
+        FUNC_GOTO_ERROR(H5E_LINK, H5E_CANTSET, FAIL, "can't set cURL username: %s", curl_err_buf);
+    if (CURLE_OK !=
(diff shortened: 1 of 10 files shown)


src/rest_vol_link.c
```diff
@@ -120,6 +121,12 @@ RV_link_create(H5VL_link_create_args_t *args, void *obj, const H5VL_loc_params_t
             new_link_loc_obj = (RV_object_t *)hard_link_target_obj;
     } /* end if */
 
+    if (new_link_loc_obj && ((base_URL = new_link_loc_obj->domain->u.file.server_info.base_URL) == NULL))
```
jhendersonHDF, line 124: These conditions need to be split so that assigning of `base_URL` doesn't depend on `new_link_loc_obj` being non-NULL

src/rest_vol_object.c
```diff
@@ -843,6 +861,33 @@ RV_object_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_object_s
                             strncpy(iter_object->u.file.filepath_name, loc_obj->u.file.filepath_name,
                                     strlen(loc_obj->u.file.filepath_name) + 1);
 
+                            if ((iter_object->u.file.server_info.username =
```
jhendersonHDF, line 864: Won't `iter_object` here already be inheriting the file/domain from `loc_obj`? If so, it seems like we'd be overwriting and leaking memory by allocating the file's `server_info` fields here

mattjala: `iter_object` has the `server_info` struct's pointers copied into it, pointing at the allocated memory of the original file object. If we didn't allocate more memory here, the server information of the original would be freed when this copy is closed.

Issue #82, opened 2023-10-09
String datatypes always use ASCII character set


mattjala:
Expected behavior: String datatypes should default to an ASCII character set, but use UTF-8 if specified in the datatype by [`H5Tset_cset`](https://docs.hdfgroup.org/hdf5/v1_12/group___a_t_o_m.html#title15)

mattjala:
Resolved in #87/#111

PR #87, opened 2023-10-13
Support UTF-8 string datatype encoding

mattjala:
Draft PR until HSDS supports fixed-length strings with UTF-8 encoding (HDFGroup/hsds#270) and this can be tested further. 

Fix for #82


mattjala:
The tests will pass once HDFGroup/hsds#278 gets in, for now they'll fail on the server version check.

jhendersonHDF (approved):

PR #95, opened 2023-11-01
Implement H5Arename/H5Arename_by_name by copy

mattjala:
This PR implements `H5Arename(_by_name)` by creating a new attribute of the desired name, copying the information, and deleting the old attribute.

The `H5Arename` test was removed from `test_rest_vol.c` because it would occasionally generate invalid datatypes for HSDS; the `H5Arename` tests in `vol-tests` are used to test this function.

### DIFF
--- a/src/rest_vol_attr.c
+++ b/src/rest_vol_attr.c
@@ -811,6 +811,7 @@ RV_attr_read(void *attr, hid_t dtype_id, void *buf, hid_t dxpl_id, void **req)
     htri_t       is_variable_str;
     size_t       dtype_size;
     size_t       host_header_len       = 0;
+    size_t       read_size             = 0;
     char        *host_header           = NULL;
     char        *url_encoded_attr_name = NULL;
     char         request_url[URL_MAX_LENGTH];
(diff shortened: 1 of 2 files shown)


src/rest_vol_attr.c
```diff
@@ -2698,6 +2874,36 @@ RV_attr_specific(void *obj, const H5VL_loc_params_t *loc_params, H5VL_attr_speci
...
+    if (renamed_attr)
+        if (RV_attr_close(renamed_attr, H5P_DEFAULT, NULL) < 0)
+            FUNC_DONE_ERROR(H5E_ATTR, H5E_CANTCLOSEOBJ, FAIL, "can't close attribute");
+
+    if (attr_parent)
+        switch (parent_obj_type) {
+            case H5I_FILE:
+            case H5I_GROUP:
+                if (RV_group_close((void *)attr_parent, H5P_DEFAULT, NULL) < 0)
+                    FUNC_GOTO_ERROR(H5E_SYM, H5E_CANTCLOSEOBJ, FAIL, "can't close parent group");
```
jhendersonHDF, line 2890: Should be FUNC_DONE_ERROR here and below so we don't potentially cause an infinite loop

jhendersonHDF (approved):
Looks fine other than the minor comment

PR #98, opened 2023-11-09
Fix bad URL when linking VOL multiple times

mattjala:
When the VOL is manually linked to multiple times, the global variable prevents connection information from being set.

It now directly checks if the connection info (base_URL) is already set.

### DIFF
--- a/src/rest_vol.c
+++ b/src/rest_vol.c
@@ -39,8 +39,7 @@
  */
 hid_t H5_rest_id_g = H5I_UNINIT;
 
-static hbool_t H5_rest_initialized_g                 = FALSE;
-static hbool_t H5_rest_connection_info_initialized_g = FALSE;
+static hbool_t H5_rest_initialized_g = FALSE;
 
 /* Identifiers for HDF5's error API */
 hid_t H5_rest_err_stack_g                 = H5I_INVALID_HID;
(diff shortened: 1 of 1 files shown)


src/rest_vol.c
```diff
@@ -729,7 +728,7 @@ H5_rest_set_connection_information(void)
     FILE             *config_file = NULL;
     herr_t            ret_value   = SUCCEED;
 
-    if (H5_rest_connection_info_initialized_g)
+    if (base_URL)
```
jhendersonHDF, line 731: Since `base_URL` is also a global, I'm surprised it doesn't have the same issues. Are we handling it differently somewhere else?

jhendersonHDF: There's also logic below dealing with freeing and replacing `base_URL` if it was already set previously. Seems like there may be some unwanted interactions with that as well.

mattjala: I think the difference is because `base_URL` is declared `extern` in `rest_vol.h`, while `H5_rest_connection_info_initialized_g` was declared statically in `rest_vol.c`.

The logic to free and replace `base_URL` doesn't make sense anymore, since setting the connection information is skipped if it's already defined. I'll remove those checks. The application won't be able to change which server it connects to during the runtime of an application, but since it can only talk to one server at a time right now anyway, I think that's probably fine. Especially since #86 will allow for different files to get different connection information in a more useful way.

jhendersonHDF (approved):

Issue #21, opened 2023-04-03
Dataset Read Does Not Support Subset of Compound Data Members


ron-kuhn:
I have a dataset consisting of a compound datatype:

- Name - 64 characters
- AcquisitionTimeUTC - I64LE
- CenterZ - F64LE
- PixelSizeZ - F64LE

I write all the rows of the dataset at one time. When I browse through the HSDS objects and download the dataset, I see that the data was written correctly. If I only want to return the Names for all the rows, it says it is successful but the 2nd Name starts with data from the 1st AcquisitionTimeUTC row.

mattjala:
Resolved in #107

PR #107, opened 2023-11-21
Handle compound type subsetting for dataset read/writes

mattjala:
This should be done for attributes too eventually, but only datasets are checked in the vol tests.

Fix for #21


mattjala:
Better implementation in #118

PR #112, opened 2023-12-18
Support link names up to 64Kb

mattjala:
Moved the URL for cURL requests to the heap to support longer URLs. 

The specific new value of `URL_MAX_LENGTH`, `((64 * 1024) + 1024) + 2048` is the maximum tested link length in the API tests (`64 * 1024 + 1024`) plus the maximum expected length of the rest of a URL (`2048`).

Requires HSDS to support long URLs, which is implemented in HDFGroup/hsds#294. This is a draft PR until that gets into master.


mattjala:
Because sending URLs longer than ~2 Kb is unreliable, this will be handled through new API exposed in HDFGroup/HSDS#296 instead.

PR #118, opened 2024-01-25
Compound subsetting through H5Tconvert and field selection

mattjala:
Alternative implementation of #107. Compound reads are handled through `H5Tconvert`. The previous issue with this - that `H5Tconvert` would overwrite the unused fields - is fixed by using the background buffer.

Compound writes are more complicated, since HSDS expects to receive a packed buffer containing only the data for the fields that are actually being written to. `H5Tconvert` leaves blank spaces in the buffer, so it was necessary to create the packed buffer through `H5Dgather` and a new callback. 

This depends on HDFGroup/HSDS#298, which fixes some issues with writing to field selections.

### DIFF
--- a/src/rest_vol_dataset.c
+++ b/src/rest_vol_dataset.c
@@ -108,6 +108,12 @@ const char *value_keys[]                  = {"value", (const char *)0};
  */
 #define BASE64_ENCODE_DEFAULT_BUFFER_SIZE 33554432 /* 32MB */
 
+/* Default separator character when specifying compounds members in a URL */
+#define COMPOUND_MEMBER_SEPARATOR ':'
+
+/* Query for server to operate on a subset of members in a compound type */
+#define COMPOUND_MEMBER_QUERY "fields="
+
 /*-------------------------------------------------------------------------
  * Function:    RV_dataset_create
  *
@@ -785,8 +791,8 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
 {
     H5S_sel_type           sel_type = H5S_SEL_ALL;
     H5T_class_t            dtype_class;
-    hbool_t                is_transfer_binary = FALSE;
-    htri_t                 contiguous         = FALSE;
+    hbool_t                is_transfer_binary  = FALSE;
+    htri_t                 is_write_contiguous = FALSE;
     htri_t                 is_variable_str;
     hssize_t               mem_select_npoints  = 0;
     hssize_t               file_select_npoints = 0;
@@ -805,6 +811,11 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
     size_t      mem_type_size  = 0;
     hbool_t     fill_bkg       = FALSE;
     const void *buf_to_write   = NULL;
+
+    hbool_t has_selection_body = FALSE;
+    char    cmpd_query[URL_MAX_LENGTH];
+    char   *member_name             = NULL;
+    char   *url_encoded_member_name = NULL;
 
     if ((transfer_info = RV_calloc(count * sizeof(dataset_transfer_info))) == NULL)
         FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't allocate space for dataset transfer info");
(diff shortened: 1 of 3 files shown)


src/rest_vol_datatype.c
```diff
@@ -2609,3 +2609,56 @@ RV_tconv_init(hid_t src_type_id, size_t *src_type_size, hid_t dst_type_id, size_
...
+    }
+
+    if ((dst_nmembs = H5Tget_nmembers(dst_type_id)) < 0)
+        FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_CANTGET, FAIL, "can't get nmembers of destination datatype");
+
+    if ((src_nmembs = H5Tget_nmembers(src_type_id)) < 0)
+        FUNC_GOTO_ERROR(H5E_DATATYPE, H5E_CANTGET, FAIL, "can't get nmembers of source datatype");
+
+    /* The library just compares the number of members to determine if two
+     * compounds are subsets, so that should suffice here as well. */
```
jhendersonHDF, line 2642: The library actually does check in H5Tconv.c(H5T__conv_struct_init) to see that all the members match up before determining the final subset status.

src/rest_vol_datatype.c
```diff
@@ -2609,3 +2609,174 @@ RV_tconv_init(hid_t src_type_id, size_t *src_type_size, hid_t dst_type_id, size_
...
+                if ((dst_member_name = H5Tget_member_name(dst_type_id, dst_idx)) == NULL)
+                    FUNC_DONE_ERROR(H5E_DATATYPE, H5E_CANTGET, FAIL,
+                                    "can't get destination datatype member name");
+
+                if ((dst_member_offset = H5Tget_member_offset(dst_type_id, dst_idx)) < 0)
+                    FUNC_DONE_ERROR(H5E_DATATYPE, H5E_CANTGET, FAIL,
+                                    "can't get destination datatype member offset");
+
+                /* Compare member names and offsets */
+                if (!strcmp(src_member_name, dst_member_name) && (src_member_offset == dst_member_offset)) {
```
jhendersonHDF, line 2671: Checking member names and offsets isn't necessarily enough since the members could be different datatypes; we also need to check if they're equal with `H5Tequal`. Unfortunate that this logic had to be re-implemented from what the library does, but maybe we should create an issue about exposing this subset detection functionality publicly for VOL connectors to use.

src/rest_vol_dataset.c
```diff
@@ -1067,13 +1121,84 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
...
+                         curl_easy_escape(curl, member_name, (int)strlen(member_name))) == NULL)
+                    FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTENCODE, FAIL,
+                                    "can't URL-encode compound member name");
+
+                if ((size_t)cmpd_query_len + strlen(url_encoded_member_name) + 1 > URL_MAX_LENGTH)
+                    FUNC_GOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL,
+                                    "compound member names too long for URL");
+
+                /* Copy name to query string without null byte */
+                memcpy(cmpd_query_ptr, member_name, strlen(member_name));
```
jhendersonHDF, line 1175: Looks like this and the reference in the line below should be `url_encoded_member_name` rather than `member_name`.

src/rest_vol_dataset.c
```diff
@@ -1067,13 +1121,84 @@ RV_dataset_write(size_t count, void *dset[], hid_t mem_type_id[], hid_t _mem_spa
...
+                cmpd_query_ptr += strlen(member_name);
+
+                /* If another member name will follow, add sepator */
+                if (j < num_cmpd_members - 1) {
+                    *cmpd_query_ptr = COMPOUND_MEMBER_SEPARATOR;
+                    cmpd_query_ptr++;
```
jhendersonHDF, line 1181: It's likely worth adding another check for going past `URL_MAX_LENGTH` here before adding the separator and below before adding the NUL terminator just to be better assured that we aren't running off the buffer end.

jhendersonHDF (approved):

PR #119, opened 2024-01-30
Support H5Fget_filesize

### LINKED ISSUE #117: Support H5Fget_filesize

mattjala:
This does not trigger a rescan on the server, so it may return a slightly out-of-date value.

Resolves #117

### DIFF
--- a/src/rest_vol_debug.c
+++ b/src/rest_vol_debug.c
@@ -439,7 +439,7 @@ file_specific_type_to_string(H5VL_file_specific_t specific_type)
  * Function:    file_optional_type_to_string
  *
  * Purpose:     Helper function to convert each member of the
- *              H5VL_file_optional_t enum into its string representation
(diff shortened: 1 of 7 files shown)


src/rest_vol_debug.c
```diff
@@ -435,6 +435,30 @@ file_specific_type_to_string(H5VL_file_specific_t specific_type)
...
+ *
+ * Return:      String representation of given object or '(unknown)' if
+ *              the function can't determine the type of object it has
+ *              been given (can't fail).
+ *
+ * Programmer:  Matthew Larson
+ *              January, 2024
+ */
+const char *
+file_optional_type_to_string(H5VL_optional_args_t optional_type)
```
jhendersonHDF, line 452: Why not just pass an H5VL_file_optional_t here so you don't need to define an enum value to determine which operation is being passed in?

jhendersonHDF (approved):

PR #124, opened 2024-02-15
Support vlen r/w with simple parent types

mattjala:
There aren't currently any tests in the API tests/vol-tests that read and write with variable length sequences, so this needs more comprehensive testing before it's ready to go in.  Point and hyperslab selection also still need to be implemented


mattjala:
- Made selection compatible with type conversion in `RV_dataset_write` by performing `H5Dgather` before `H5Tconvert` when the selection is noncontiguous.
- Refactored the write pipeline to use a unique buffer for each step that uses dynamic allocation: `gather_buf` for gathering selections, `selection_buf` for putting information about point selections into the body, `base64_encoded_values` for the encoded values during point selection,`tconv_buf`/`bkg_buf` for type converison, and `serialize_buf` for serialization to json/binary. `uinfo.buffer` is updated after each step to always points at the buffer containing the processed data for the write.

mattjala:
- Variable length I/O is now done through binary rather than JSON for non-point selections
- HSDS doesn't accept point selections as a request parameters, so point selection is still done through JSON
- Variable length point selection depends on HDFGroup/hsds#325

src/rest_vol.c
```diff
@@ -3794,11 +3796,60 @@ RV_curl_multi_perform(CURL *curl_multi_handle, dataset_transfer_info *transfer_i
...
                     if (transfer_info[handle_index].transfer_type == WRITE) {
-                        RV_free(transfer_info[handle_index].u.write_info.write_body);
-                        transfer_info[handle_index].u.write_info.write_body = NULL;
+                        if (transfer_info[handle_index].tconv_buf) {
+                            if ((dtype_class = H5Tget_class(transfer_info[handle_index].mem_type_id)) ==
+                                H5T_NO_CLASS)
+                                FUNC_DONE_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get mem dtype class");
+
+                            /* Clean up memory allocated by type conversion of vlen types */
+                            if (dtype_class == H5T_VLEN) {
```
jhendersonHDF, line 3805: This may become complicated for things like compound datatypes with variable-length fields inside. You probably want to use `H5Tdetect_class(..., H5T_VLEN)` for this.

src/rest_vol_dataset.c
```diff
@@ -3991,15 +4039,29 @@ RV_convert_dataspace_selection_to_string(hid_t space_id, char **selection_string
...
+#endif
+                /* Format the point selection as a binary buffer containing each coordinate index as an
+                 * hsize_t */
+                if ((num_points = H5Sget_select_npoints(space_id)) < 0)
+                    FUNC_GOTO_ERROR(H5E_DATASPACE, H5E_CANTGET, FAIL, "can't get number of selected points");
+
+                point_buf_size = (size_t)(ndims * num_points) * sizeof(hsize_t);
+                CHECKED_REALLOC_NO_PTR(out_string, out_string_len, point_buf_size, H5E_DATASPACE, FAIL);
+
+                if (H5Sget_select_elem_pointlist(space_id, 0, (hsize_t)num_points, (hsize_t *)out_string) < 0)
```
jhendersonHDF, line 4055: Have you tested point selections to verify that this approach works? Whereas for hyperslabs the selection is actually converted into a string, this appears to be just inserting `hsize_t` values into a `char *` buffer, which is not really what this function is aiming to do. It's also not correct to cast a `char *` (`out_string`) to `hsize_t *` here due to alignment requirements.

At the very least, if you're really just wanting a binary buffer of the `hsize_t` values for each point coordinate, that functionality should like be in a separate function to make it clear that we aren't actually expecting a string back.

src/rest_vol_dataset.c
```diff
@@ -5013,8 +5078,26 @@ RV_json_values_to_binary_callback(char *HTTP_response, const void *callback_data
...
+
+        if ((dtype_class = H5Tget_class(dtype_id)) < 0)
+            FUNC_DONE_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get datatype class");
+
+        if (dtype_class == H5T_VLEN) {
+            /* Reclaim each allocated vlen buffer */
+            hvl_t *curr = (hvl_t *)value_buffer;
+
+            while (curr->p) {
+                RV_free(curr->p);
```
jhendersonHDF, line 5092: Another case here where `H5Treclaim` is probably simpler and more correct than trying to do this by hand.

src/rest_vol_dataset.c
```diff
@@ -5023,86 +5106,155 @@ RV_json_values_to_binary_callback(char *HTTP_response, const void *callback_data
...
+            if ((vl->p = calloc(vl->len, parent_dtype_size)) == NULL)
+                FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL,
+                                "can't allocate space for returned vlen seq data");
+
+            /* Recursively encode each element in the sequence and place it in the allocated buffer */
+            for (size_t i = 0; i < vl->len; i++) {
+                if ((seq_elem_val = YAJL_GET_ARRAY(value_entry)->values[i]) == NULL)
+                    FUNC_GOTO_ERROR(H5E_DATASET, H5E_PARSEERROR, FAIL, "can't parse vlen sequence element");
+
+                seq_elem_ptr = (void *)(((char *)vl->p) + i * parent_dtype_size);
```
jhendersonHDF, line 5232: Fine for now, though it seems this would only work for simple cases. For example, if the parent datatype is also a vlen I would expect the calculations here for the next element to not be correct. The function would likely need to be able to pass the current offset into the buffer both down and back up.

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get size of vlen seq base type");
+
+    if ((out_string = (char *)calloc(DATASET_VLEN_JSON_BODY_DEFAULT_SIZE, 1)) == NULL)
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't allocate space for vlen data JSON");
+
+    *out_size           = DATASET_CREATION_PROPERTIES_BODY_DEFAULT_SIZE;
+    out_string_curr_pos = out_string;
+
+    /* Delineate start of list of sequences */
+    out_diff     = out_string_curr_pos - out_string;
```
jhendersonHDF, line 5432: `out_string_curr_pos` is set to `out_string` directly above, so this and `tgt_out_size` below are always going to be 0 here. I don't think there's any need for these lines or the `CHECKED_REALLOC`.

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+ *              Allocates memory under the *out pointer on success which must be
+ *              freed by the calling function.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  Matthew Larson
+ *              January, 2024
+ */
+static herr_t
+RV_vlen_data_to_json(const void *_in, size_t nelems, hid_t dtype_id, void **out, size_t *out_size)
```
jhendersonHDF, line 5389: Is `_in` meant to be anything other than a `const hvl_t *`? If not, just change this from `const void *` to `const hvl_t *` to get a little more type safety and remove the unnecessary casting below. Avoid passing `void *` anywhere that it isn't necessary.

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+        /* Delineate start of this sequence */
+        out_diff     = out_string_curr_pos - out_string;
+        tgt_out_size = (size_t)(out_diff) + 1;
+        CHECKED_REALLOC(out_string, *out_size, tgt_out_size, out_string_curr_pos, H5E_DATASET, FAIL);
+        out_string = strcat(out_string, "[");
+        out_string_curr_pos += 1;
+
+        /* Stringify each element in a second fixed-size buffer, then copy to sequence buffer */
+        for (size_t j = 0; j < seq.len; j++) {
+#define MAX_SEQ_ELEM_BUF_SIZE 128
```
jhendersonHDF, line 5452: We should generally avoid inline `#define`s anywhere and keep them either at the top of the file or at the very least right above the function they're used in. That said, it doesn't seem like we necessarily need to have a fixed limit on the number of sequence elements we support, but if so, we should at least throw an error if a sequence is too large.

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+            char  seq_elem_buffer[MAX_SEQ_ELEM_BUF_SIZE];
+            char *seq_elem_buffer_ptr = seq_elem_buffer;
+
+            memset(seq_elem_buffer, 0, MAX_SEQ_ELEM_BUF_SIZE);
+
+            switch (parent_class) {
+                case (H5T_INTEGER):
+                    if ((bytes_printed = snprintf(seq_elem_buffer_ptr, MAX_SEQ_ELEM_BUF_SIZE, "%d",
```
jhendersonHDF, line 5460: This should at least verify that the parent datatype matches `H5T_NATIVE_INT` before trying to cast the buffer to `int *` and retrieve an element from it. Ideally, we should fully support the integer types, but I understand that's a lot more complicated than just supporting the native types.

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+                                                  ((int *)seq.p)[j])) < 0)
+                        FUNC_GOTO_ERROR(H5E_DATASET, H5E_SYSERRSTR, FAIL, "snprintf error");
+
+                    seq_elem_buffer_ptr += bytes_printed;
+                    break;
+                case (H5T_FLOAT):
+                    if ((bytes_printed = snprintf(seq_elem_buffer_ptr, MAX_SEQ_ELEM_BUF_SIZE, "%f",
```
jhendersonHDF, line 5467: Same here about verifying against `H5T_NATIVE_FLOAT` before casting the buffer to `float *`.

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+    if ((parent_class = H5Tget_class(parent_type)) < 0)
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get class of parent of vlen dtype");
+
+    if ((parent_dtype_size = H5Tget_size(parent_type)) == 0)
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get size of vlen seq base type");
+
+    if ((out_string = (char *)calloc(DATASET_VLEN_JSON_BODY_DEFAULT_SIZE, 1)) == NULL)
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't allocate space for vlen data JSON");
+
+    *out_size           = DATASET_CREATION_PROPERTIES_BODY_DEFAULT_SIZE;
```
jhendersonHDF, line 5428: Looks like this should be `DATASET_VLEN_JSON_BODY_DEFAULT_SIZE`

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_UNSUPPORTED, FAIL,
+                        "vlen datatypes are only supported for int, float, and string base types");
+
+    if ((parent_dtype_size = H5Tget_size(parent_dtype)) == 0)
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get size of vlen parent datatype");
+
+    if ((out_buf = calloc(DATASET_VLEN_JSON_BODY_DEFAULT_SIZE, 1)) == NULL)
+        FUNC_GOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't allocate space for packed vlen data");
+
+    *out_size        = DATASET_CREATION_PROPERTIES_BODY_DEFAULT_SIZE;
```
jhendersonHDF, line 5610: Looks like this should be `DATASET_VLEN_JSON_BODY_DEFAULT_SIZE`

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+ *              Allocate memory under the *out pointer on success which must
+ *              be freed by calling function.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  Matthew Larson
+ *              March, 2024
+ */
+static herr_t
+RV_pack_vlen_data(const void *_in, size_t nelems, hid_t dtype_id, void **out, size_t *out_size)
```
jhendersonHDF, line 5570: Same comment here about just changing this to be `const hvl_t *`

src/rest_vol_dataset.c
```diff
@@ -5217,6 +5369,370 @@ RV_get_cmpd_subset_nmembers(hid_t src_type_id, hid_t dst_type_id, size_t *num_cm
...
+    }
+
+    *out = out_buf;
+
+done:
+
+    if (ret_value < 0 && out_buf) {
+        /* Free any memory allocated for individual sequences */
+        for (size_t i = 0; i < nelems; i++) {
+            vl = ((hvl_t *)out_buf) + i;
```
jhendersonHDF, line 5722: Casting `out_buf` to `hvl_t *` here is not correct due to alignment requirements. You should be able to find examples of this elsewhere, but the proper way is to have a local `hvl_t` stack variable that you memcpy into from the appropriate index into `out_buf` (as a `char *`) and then perform your free operations on that variable's `.p` field.

In this case no modifications to the buffer need to be made since it was allocated locally, but if you needed to modify something in the original buffer (such as when setting `vl->p` to NULL here), you'd memcpy from the stack variable back into the buffer at the same index and then move to the next element index.

jhendersonHDF (approved):

PR #127, opened 2024-04-15
Add Windows CI and build script

### LINKED ISSUE #116: Test support for Windows in CI

mattjala:
- Added workflow to build and test on Windows with CMake
- Populated `build_vol_cmake.bat` to replicate `build_vol_cmake.sh`
- Various CMake changes so that YAJL and cURL can be discovered on Windows. The github runners don't seem to have a FindCURL module, so the CMake now defines `CURL_LIBRARIES` and `CURL_INCLUDE_DIRS` relative to the provided `CURL_ROOT`.

Changes to the VOL for Windows compatibility:
- Replaced `clock_gettime` with `RV_now_usec`, which uses one of `clock_gettime`, `gettimeofday`, or `time`.
- Replaced strtok_r with cross-platform strtok

Resolves #116


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # CMake Build Script for the HDF5 REST VOL connector #
 #----------------------------------------------------#
 
-cmake_minimum_required (VERSION 3.10)
+cmake_minimum_required (VERSION 3.10...3.29)
```
jhendersonHDF, line 5: I don't think we should be setting a maximum version here, as it would then need to be maintained as new versions of CMake are released. Otherwise, it could result in surprising behavior down the line.

CMakeLists.txt
```diff
@@ -118,12 +117,27 @@ if (HDF5_ENABLE_THREADSAFE AND NOT Threads_FOUND)
...
 
+set (CURL_TARGET_LIBTYPE shared)
+
+
 find_package (CURL 7.61 REQUIRED)
 if (CURL_FOUND)
+  # FindCURL module is not always found on windows; set paths manually
+  if (WIN32)
+    set (CURL_INCLUDE_DIRS "${CURL_ROOT}\\include")
+    set (LINK_LIBS ${LINK_LIBS} Ws2_32.lib Wldap32.lib)
```
jhendersonHDF, line 128: We shouldn't ever manually work around problems with a "Find" module like this. The issue is likely that cmake on those runners isn't new enough. Since FindCURL was introduced in version 3.12 of cmake, we should make that the minimum version for `cmake_minimum_required`.

src/rest_vol.c
```diff
@@ -3487,23 +3486,23 @@ RV_parse_server_version(char *HTTP_response, const void *callback_data_in, void
         FUNC_GOTO_ERROR(H5E_OBJECT, H5E_BADVALUE, FAIL, "server version was NULL");
 
     /* Parse server version into struct */
-    if (NULL == (version_field = strtok_r(version_response, ".", &saveptr)))
+    if (NULL == (version_field = strtok(version_response, ".")))
```
jhendersonHDF, line 3489: I'd need to consult the YAJL docs, but it's possible this isn't safe since strtok modifies its input. MSVC is also generally POSIX-compliant for newer versions and should have either `strtok_r` or `strtok_s` available (see H5win32defs.h in HDF5).

CMake/FindYAJL.cmake
```diff
@@ -47,10 +47,12 @@ if (YAJL_FOUND)
   endif ()
   set_target_properties(yajl-shared PROPERTIES
     IMPORTED_LOCATION "${YAJL_SHARED_LIBRARY}"
+    IMPORTED_IMPLIB "${YAJL_SHARED_LIBRARY}"
     INTERFACE_INCLUDE_DIRECTORIES "${YAJL_INCLUDE_DIR}"
   )
   set_target_properties(yajl-static PROPERTIES
     IMPORTED_LOCATION "${YAJL_STATIC_LIBRARY}"
+    IMPORTED_IMPLIB "${YAJL_SHARED_LIBRARY}"
```
jhendersonHDF, line 55: `YAJL_SHARED_LIBRARY` -> `YAJL_STATIC_LIBRARY`

CMakeLists.txt
```diff
@@ -368,6 +378,10 @@ if (NOT BUILD_SHARED_LIBS)
 endif ()
 if (BUILD_SHARED_LIBS)
   set (HDF5_VOL_REST_LIBRARIES_TO_EXPORT ${HDF5_VOL_REST_LIBRARIES_TO_EXPORT} ${HDF5_VOL_REST_LIBSH_TARGET} CACHE INTERNAL "Store which libraries should be exported" FORCE)
+  # Value is not automatically inherited from cache on Windows
```
jhendersonHDF, line 381: This and the same below must either be a CMake bug or something wrong in the usage of this variable in the CMake code somewhere. CMake behavior doesn't generally change across platforms, except for things like quoting and other edge cases. The set call above should be enough for any case needed, though it is a bit heavy-handed.

CMakeLists.txt
```diff
@@ -118,12 +116,16 @@ if (HDF5_ENABLE_THREADSAFE AND NOT Threads_FOUND)
   find_package(Threads)
 endif ()
 
-find_package (CURL 7.61 REQUIRED)
+set (CURL_TARGET_LIBTYPE shared)
+
+set(CURL_NO_CURL_CMAKE ON)
```
jhendersonHDF, line 121: This seems like something that really should only be set for Windows and not other platforms

jhendersonHDF: That said, it's mentioned that curl is often built with cmake these days anyway, so setting this to ON shouldn't necessarily be needed in the first place

build_vol_cmake.sh
```diff
@@ -119,7 +121,11 @@ while getopts "$optspec" optchar; do
         echo
         ;;
     t)  YAJL_LIB_OPT="-DYAJL_USE_STATIC_LIBRARIES=ON"
-        echo "Use the static YAJL library."
+        echo "Using the static YAJL library."
+        echo
+        ;;
+    c)  CMAKE_C_FLAGS="/DCURL_STATICLIB"
```
jhendersonHDF, line 127: Should the above be changed or should this be changed or this intended? There seems to be a discrepancy here between using CMake -D-style options or passing the option directly in via `CMAKE_C_FLAGS`

mattjala: This should use the regular CMake style options. It was copied from the windows build script which [requires preprocessor definitions to be used](https://curl.se/docs/faq.html#Link_errors_when_building_libcur).

jhendersonHDF (approved):

PR #130, opened 2024-11-18
Support running tests against HDF5 2.0.0 with CMake builds

mattjala:
- If the REST VOL is built against HDF5 >= 2.0.0, update the random datatype generator to avoid generating complex datatypes
- Disable running tests with autotools against HDF5 develop, since HDF5 >= 2.0.0 doesn't support autotools builds


configure.ac
```diff
@@ -1161,6 +1161,9 @@ else
   enable_static=no
 fi
 
+# Autotools build isn't supported by HDF5 >= 2.0.0, so assume HDF5 version is < 2.0.0
+AC_DEFINE([HDF5_2], [0], [Whether HDF5 version is at least 2.0.0])
```
jhendersonHDF, line 1165: Since Autotools build are still supported in develop, I think we should keep testing until the Autotools code is removed

jhendersonHDF (approved):

PR #131, opened 2024-12-13
Fix dataset creation failure on Ubuntu 24

mattjala:
- Fix bad length value being provided to snprintf

A pointer value in `RV_convert_dataspace_shape_to_JSON` was cast to size_t in a way that didn't make sense, resulting in snprintf being provided an essentially random maximum length. For some reason, this only caused problems on Ubuntu 24.

- Explicitly add Ubuntu 22.04 to CI

A recent update to the github runners changed ubuntu-latest from 22.04 to 24.04.

- Merge autotools/cmake CI jobs into one using matrix to reduce duplication

- Remove vol-tests submodule

As of hdf5#5170, the vol-tests don't successfully build anymore. These haven't been used for testing since the API tests were moved back into the library, so it makes sense to remove them.

- Fix several type size warnings regarding `hsize_t`


src/rest_vol.c
```diff
@@ -4125,8 +4124,7 @@ RV_curl_delete(CURL *curl_handle, server_info_t *server_info, const char *reques
 
     strcpy(host_header, host_string);
 
-    curl_headers_local = curl_slist_append(
-        curl_headers_local, strncat(host_header, filename, host_header_len - strlen(host_string) - 1));
```
jhendersonHDF, line 4129: Why were these converted back to `strcat` calls?

mattjala: The lengths provided to `strncat` were pulled from the length of the target strings anyway, so having these be `strncat` didn't really accomplish anything.

jhendersonHDF: Hmm, in this case yes they weren't doing much, though they're more future-proof than a regular `strcat` in case someone adds more things to the header in the future and forgets to allocate the space correctly. Really, I think most of these should just be switched to a single `snprintf` anyway to avoid most of the problems of having to manage the buffer size.

.github/workflows/main.yml
```diff
@@ -59,32 +72,78 @@ jobs:
...
+      - name: CMake Configure + Build HDF5
+        if: matrix.build_system == 'cmake'
+        run: |
+          mkdir ${{github.workspace}}/hdf5/build
+          cd ./build
+          cmake \
+          -DHDF5_BUILD_HL_LIB=ON \
+          -DBUILD_SHARED_LIBS=ON -DHDF5_ENABLE_SZIP_SUPPORT=OFF \
+          -DHDF5_TEST_API=ON \
+          -DHDF5_ENABLE_Z_LIB_SUPPORT=OFF \
```
jhendersonHDF, line 84: Just a note, depending on which branch you're building this option could either be `Z_LIB` or `ZLIB` now. Though it shouldn't be a problem for develop as long as the default stays OFF

.github/workflows/main.yml
```diff
@@ -140,146 +198,27 @@ jobs:
...
-      - name: Install valgrind
-        run: |
-          sudo apt update
-          sudo apt install valgrind
-        working-directory: ${{ github.workspace }}
-
-        # Requests 2.32.0 breaks requests-unixsocket, used by HSDS for socket connections
-      - name: Fix requests version
+      - name: Test REST VOL (CMake)
+        if: matrix.build_system == 'cmmake'
```
jhendersonHDF, line 212: `cmmake` -> `cmake`

.github/workflows/main.yml
```diff
@@ -140,146 +198,27 @@ jobs:
...
-            mkdir ${{github.workspace}}/hsdsdata &&
-            mkdir ${{github.workspace}}/hsdsdata/hsdstest &&
-            cp admin/config/groups.default admin/config/groups.txt &&
-            cp admin/config/passwd.default admin/config/passwd.txt &&
-            cp admin/config/groups.default admin/config/groups.txt &&
-            cp admin/config/passwd.default admin/config/passwd.txt
-            ROOT_DIR=${{github.workspace}}/hsdadata ./runall.sh --no-docker 1 &
-            sleep 10
-        working-directory: ${{github.workspace}}/hsds
+              valgrind --leak-check=full -s ctest -R "test_rest_vol" -VV
```
jhendersonHDF, line 215: There's no real reason to use valgrind here since `ctest` will be wrapping running the tests. We also shouldn't generally use valgrind in the CI unless there's a specific reason

jhendersonHDF (approved):

PR #134, opened 2026-07-27
Replace yajl with vendored yyjson for JSON parsing

mattjala:
The REST VOL depended on the unmaintained lloyd/yajl library to parse HSDS's JSON responses. Its ancient CMakeLists (cmake_minimum_required 2.6) fails under CMake 4.x, which [broke the Windows CI job](https://github.com/HDFGroup/vol-rest/actions/runs/30041808993) that builds yajl from source.

This PR entirely replaces lloyd/yajl with the actively-maintained yyjson library. More specifically:

- Add JSON helpers in rest_vol.c/.h using yyjson: RV_json_parse, RV_json_get, RV_json_obj_key_at/RV_json_obj_val_at, and RV_json_get_*/RV_json_is_* accessors.
- Port all yajl usage to the new routines.

Some associated build changes:
- CMake: drop find_package(YAJL) and the yajl link, add yyjson.c to sources, remove the unused CMake/FindYAJL.cmake, its install rule, and the HDF5_VOL_REST_RESOURCES_MOD_DIR variable.
- Autotools: add yyjson.c to src/Makefile.am, remove -lyajl, the YAJL install-dir flags, the libyajl check and --with-yajl from configure.ac.


src/yyjson.c
jhendersonHDF, line 1: Just as a first comment, we should avoid vendoring any dependencies where we can since it will just result in maintenance burden. yyjson appears to ship a CMake .config file, so a simple `find_package()` call should be enough to setup what we need.

mattjala: Sure. Are we still concerned with supporting autotools builds for the REST VOL, or can we drop it since HDF5 has?

